PR #1, opened 2019-02-01
update to FCS 26.0.1 to allow FSAC to update as well

baronfel:
Need to update FCS dependency so that FSAC is able to update the FCS dependency.


Krzysztof-Cieslak:
Thanks!

baronfel:
Hey! Could you push to nuget so I can flow through the rest of the FCS update?

Krzysztof-Cieslak:
Package released, it should be online after NuGet does it magic

baronfel:
you rock

PR #7, opened 2019-12-16
Add CLI tool to run analyzers out of editor

Krzysztof-Cieslak:
This is still WIP, but it's already working. 

CC: @baronfel & @TheAngryByrd


TheAngryByrd:
Definitely feel free to steal anything from https://github.com/TheAngryByrd/FSharp.Analyzers/tree/master/src/FSharp.Analyzer.App

TheAngryByrd:
I think similar to how MSBuild operates, we'd like to have the [Fail on warnings](https://github.com/TheAngryByrd/FSharp.Analyzers/blob/master/src/FSharp.Analyzer.App/Commands/AnalyzeCommand.fs#L19) be something we can also pass in as a list so we get a proper exit code when we hit something with that severity.

TheAngryByrd:
We probably also want to consider ways to disable analyzing for code sections or ignore files.  FSharpLint has a convention for this: https://fsprojects.github.io/FSharpLint/

Krzysztof-Cieslak:
I've added:
* Fail on warnings
* Ignore files
* Verbose mode
* Colorized output

I'm pretty happy with current state, I think

src/FSharp.Analyzers.Cli/Program.fs
```diff
@@ -0,0 +1,186 @@
...
+            | ProjectResponse.Project proj ->
+                printInfo "Project %s loaded" file
+                proj.projectFiles
+                |> List.choose (fun file ->
+                    projectSystem.GetProjectOptions file
+                    |> Option.map (fun opts -> file, opts)
+                )
+            | ProjectResponse.ProjectError(errorDetails) ->
+                printError "Project loading faield: %A" errorDetails
+                []
```
TheAngryByrd: Do you think the tool should exit with a non-zero exit code if we can't load the project?

src/FSharp.Analyzers.Cli/Program.fs
```diff
@@ -0,0 +1,186 @@
...
+    if check then -12345 else 0
+
+[<EntryPoint>]
+let main argv =
+    let results = parser.ParseCommandLine argv
+    verbose <- results.Contains <@ Verbose @>
+    printInfo "Running in verbose mode%s" ""
+
+    let failOnWarnings = results.GetResult(<@ Fail_On_Warnings @>, [])
+    printInfo "Fail On Warnings: %A" failOnWarnings
```
TheAngryByrd: Depending on how many warnings `%A` might not print all of them.  I think it's worth considering a `String.join` here.

src/FSharp.Analyzers.Cli/Program.fs
```diff
@@ -0,0 +1,186 @@
...
+
+    let ignoreFiles = results.GetResult(<@ Ignore_Files @>, [])
+    printInfo "Ignore Files: %A" ignoreFiles
```
TheAngryByrd: Depending on how many warnings `%A` might not print all of them.  I think it's worth considering a `String.join` here.

src/FSharp.Analyzers.Cli/Program.fs
TheAngryByrd: Also typo `faield`

Zaid-Ajaj:
Analyzer resolution seems a bit hidden away in this implementation where the CLI is looking for directory "analyzers": it is not clear (unless I missed something) how the end-user should download analyzers and put them there. Isn't it a better idea to use the directory `{cwd}/packages/analyzers` instead of `{cwd}/analyzers` so that the end-user can use paket and include the analyzer nuget packages within a `Analyzers` dependency group (paket `storage` option cannot be `none` for this group)?

src/FSharp.Analyzers.Cli/Program.fs
```diff
@@ -0,0 +1,186 @@
...
+open FSharp.Compiler.SourceCodeServices
+open FSharp.Compiler.Text
+open ProjectSystem
+open Argu
+open FSharp.Analyzers.SDK
+open GlobExpressions
+
+type Arguments =
+    | Project of string
+    | Analyzer of string
```
Zaid-Ajaj: Suggestion: rename to `Analyzers_Path` since this is the path of the directory in which a bunch analyzers can be loaded?

Krzysztof-Cieslak:
> Isn't it a better idea to use the directory `{cwd}/packages/analyzers` instead of `{cwd}/analyzers`

Oh, yes, I was supposed to change default path to `packages/analyzers` just forgot about that 😁

Krzysztof-Cieslak:
Applied all the feedback.

src/FSharp.Analyzers.Cli/Program.fs
Krzysztof-Cieslak: Done

PR #13, opened 2020-02-18
Docs: Writing Analyzers, Requirements, Packaging and Distribution

Zaid-Ajaj:
I have been fighting with packing and distribution of my analyzer yesterday, so I though I could write a little bit of docs for people who want to write them and for the future me who will eventually forget about how he solved the problem :smile:

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,91 @@ Library used for building custom analyzers for FSAC / F# editors.
 ## How to run sample
 1. `dotnet fake build -t RunSample`
 
+### Writing Analyzers
+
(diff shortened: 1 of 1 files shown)


Zaid-Ajaj:
Note: I searched the dotnet docs on how to publish the assemblies of third-party dependencies using just `dotnet pack` and all links say that I should add
```xml
<CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies>
```
in a `PropertyGroup` but that didn't work, nor did the `copy_local: true` trick from paket so I went and implemented the copying myself which is the documented FAKE target

Krzysztof-Cieslak:
Awesome, thanks

Issue #8, opened 2020-01-18
Feature: Passing options through the context to named analyzers


Zaid-Ajaj:
**Is your feature request related to a problem? Please describe.**
The end user does not have any way to control the behavior the used analyzer within a project. Just like with code linters, it would great if the user could provide options for the analyzers on a per-project basis. One use-case would be to control the strictness of an analyzer and changing the severity of certain errors from compile errors into warning. 

**Describe the solution you'd like**
For this to work, the SDK must have a way to *identify* the used analyzers. One way to do that is by giving analyzers an optional name from the `Analyzer` attribute:
```fsharp
[<Analyzer "MyAnalyzerName">]
```
Then with the help of ionide, the SDK would pick up a configuration file called `fsharp-analyzers.json` that has the following structure:
```json
{
  "MyAnalyzerName": {
     "Mode": "Strict",
     "ErrorsAsWarnings": "true" 
  }, 

  "OtherAnalyzer": { 
     "ErrorsAsWarnings": "true" 
   }
}
```
The SDK can then read this JSON file and parse the options as `Map<string, string>` per analyzer where this `Map<string, string>` would be available through the context input of the analyzer. For example, for the "MyAnalyzerName", the options would be
```fsharp
Map.ofList [ "Mode", "Strict"; "ErrorsAsWarnings", "true" ]
```

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

cmeeren:
I am implementing an F# linter. This is an absolute must-have for me if I am to implement it using FSharp.Analyzers.SDK (which also depends on some other aspects).

I think that ideally, configs should be searched for in the current analyzed file's directory and all parent directories, with deeper levels overriding settings from higher levels. Perhaps `Microsoft.Extensions.Configuration` can be used for a lot of the boilerplate logic here (it's what's used for `appsettings.json` in e.g. ASP.NET Core projects, but is not dependent on ASP.NET Core). Note that just like `.editorconfig`, the config should support a top-level `isRoot` flag where no higher-level files are considered (to avoid your linter settings being inadvertently modified by configs outside your repo root).

Note that depending on how an analyzer library implements analyzers, the analyzers may need to be namespaced using e.g. the analyzer assembly name in the config and elsewhere, to avoid conflicts.

cmeeren:
[...]

PR #16, opened 2020-03-08
Named analyzers and better error handling

Zaid-Ajaj:
When the SDK runs analyzers, it swallows all errors making it impossible for FSAC to know which analyzer is problematic and makes troubleshooting harder. This PR addresses this issue by allowing the analyzers to be (optionally) named via the attribute like this `[<Analuzer "AnalyzerName">]`. I added a new function to run the analyzers of this signature:
```fs
val runAnalyzersSafely : Context -> AnalysisResult list
```
Where each `AnalysisResult` is the result of each registered analyzer and is defined as follows:
```fs
type AnalysisResult = {
  AnalyzerName : string
  Output : Result<Message list, exn>
}
```
This way, we will be able to identify which analyzer wasn't able to run successfully and why (cc @baronfel) 

This is also preparation for fixing #8 which required the analyzers to be named.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,26 +22,34 @@ module BadCodeAnalyzer
 open FSharp.Analyzers.SDK
 
 [<Analyzer>]
-let badCodeAnalyzer : Analyzer = 
-  fun (context: Context) = 
+let badCodeAnalyzer : Analyzer =
+  fun (context: Context) ->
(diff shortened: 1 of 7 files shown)


Krzysztof-Cieslak:
Thanks.

PR #21, opened 2020-05-01
Fixed column-line order in print, also fixes Ionide code highlight.

Thorium:
Very small PR. 
Didn't broke CI, there was a problem already !


Krzysztof-Cieslak:
I don't think it fixed Ionide code highlighting as `FSharp.Analyzers.Cli/Program.fs` is just CLI part, but thanks for the fix! :-)

PR #31, opened 2020-12-23
Add support for returning tooltips

Krzysztof-Cieslak:
So this basically enhances our API with the ability to return tooltips (which is just range and text). The client (like FSAC) will need to put returned tooltips in its internal state/cache and then when the user requests a tooltip at the position it will need to query its state/cache, find if there's any Analyzer Tooltip on given position and append a message to normal tooltip it would return at this position. 

@dsyme, I know you have some need for tooltips in analyzers, anything, in particular, you'd need?


tforkmann:
This would be great!

src/FSharp.Analyzers.SDK/FSharp.Analyzers.SDK.fs
```diff
@@ -39,4 +39,15 @@ type Message =
       Range: Range.range
       Fixes: Fix list }
 
-type Analyzer = Context -> Message list
\ No newline at end of file
+type Tooltip =
+    { Message: string
```
dsyme, line 43: Is this "Message" in HTML, or markdown, or ...?

dsyme:
This looks great.  I am planning on using it.

BTW when running a particular version of Ionide in VSCode what's the right way to determine the corresponding version of FsAutoComplete, FSharp.Analyzers.SDK and FCS ?

PR #40, opened 2021-11-13
FSharp.Compiler.Service 41.0.1

TheAngryByrd:
Getting analyzers on the latests FSharp.Compiler.Service


src/FSharp.Analyzers.Cli/Program.fs
```diff
@@ -23,7 +23,7 @@ let mutable verbose = false
 
 let createFCS () =
     let checker = FSharpChecker.Create(projectCacheSize = 200, keepAllBackgroundResolutions = true, keepAssemblyContents = true)
-    checker.ImplicitlyStartBackgroundWork <- true
+    // checker.ImplicitlyStartBackgroundWork <- true
```
TheAngryByrd, line 26: `ImplicitlyStartBackgroundWork` no longer exists.  Not sure what action to take here.

baronfel: That's fine - it was part of removing the reactor thread, so it's safe to remove

PR #41, opened 2022-04-17
update for net6.0 and dependencies

jcmrva:


.gitignore
```diff
@@ -8,4 +8,5 @@ out
 AssemblyInfo.fs
 .DS_Store
 release.cmd
-analyzers
\ No newline at end of file
+analyzers
+.ionide
```
baronfel, line 12: with the most recent Ionide this shouldn't be required anymore - does the `.ionide` directory spawn with Ionide 6? (you'll have to delete it if it existed beforehand)

global.json
```diff
@@ -1,5 +1,6 @@
 {
     "sdk": {
-        "version": "5.0.402"
+        "version": "6.0.100",
```
baronfel, line 3: mind bumping this to 6.0.200 at least? that's the most recent SDK feature band.

baronfel (changes requested):

.gitignore
jcmrva: It doesn't, that was added via extension suggestion.
> You are missing entries in your .gitignore for Ionide-specific data files. Would you like to add them?

baronfel: oh lol, guess I can update that check in ionide itself now that it's no longer necessary.

baronfel (approved):

baronfel:
Thanks for this!

PR #42, opened 2022-04-25
include the full license file

jcmrva:


baronfel:
What's driving this @jcmrva? I tend to prefer the SPDX expressions because tools know how to parse them.

jcmrva:
Sorry, I should've commented because I no longer recall what prompted this. Something I read... 🤷

cmeeren:
This may have been promoted by #42. But I think it can be solved by using `<PackageLicenseExpression>MIT</PackageLicenseExpression>` in the tool project. I have it on my to-do list of housekeeping changes.

PR #50, opened 2022-12-29
Modernize tooling

### LINKED ISSUE #49: Modernize/simplify tooling/CI/packaging?
As may be evident from recent issues and comments, I'd like to help out in the development if FSharp.Analyzers.SDK. In that light, I am wondering if you are open to changes that I think would make it simpler to contribute and make at least me feel more at home. 

I am willing to make any or all of these changes myself through PRs.

In short, I would like to be able to open the solution in VS/Rider/VSCode and have it "just work" using standard .NET tooling, which is really good nowadays (better than some years ago). Moreover, I would like to not have the additional complexity of Fake and Paket to think about.

These suggestions are based on several years of experience with several other F# OSS packages I have created and maintain, as well as tens of F# projects at work.

I realize these may be controversial suggestions and fully expect to be turned down, but given the lack of activity in this repo (last commit over 8 months ago), the current maintainers may want to lower the barrier of entry for contibutions or even bring on additional maintainers, so I'd like to ask anyway.

* Use NuGet (with lock files) instead of Paket. I see no compelling reason to use Paket now that NuGet has become so good in recent years, with support for lock files. NuGet is more compatible with stock tooling (e.g. VS), and is the "default" for the .NET ecosystem.
* Remove Fake, unless it does something that is difficult to accomplish otherwise. When it comes to packaging and releasing, I have yet to experience problems with stock GitHub and `dotnet` CLI functionality, assembly/package information in project files, etc. For releases, I keep a manually curated `RELEASE_NOTES.md` and have always found this more than adequate, and if I need more documentation than fits in the readme, I have used a separate easily searchable `DOCUMENTATION.md` file like [Giraffe does](https://github.com/giraffe-fsharp/Giraffe/blob/master/DOCUMENTATION.md).
* Install Fantomas as a dotnet tool and reformat the code, then enforce code style during CI. I do this in Flinter with [this config](https://github.com/cmeeren/Flinter/blob/e08899d4903f2d75c431d37eb71b27d157bc9e70/.editorconfig), which is basically the standard except Stroustrup brackets are enabled (which I find gives more readable code that is easier to edit and has less line damage in diffs).
* Build on GitHub. Why use an external service when everything you need is right here? (I realize that may not have been the case when the project was set up, of course.)
* Set up a `Directory.Build.props` that looks like [this](https://github.com/cmeeren/Flinter/blob/e08899d4903f2d75c431d37eb71b27d157bc9e70/Directory.Build.props). It uses DotNet.ReproducibleBuilds, enables warnings as errors, and warns (errors) on unused values.
* Add a strict `global.json` to ensure repeatable builds (minor versions should be compatible, but bugs in FSharp.Core packaging have broken my builds even on minor updates to dotnet tooling on build agents)
* Add `nuget.config` to ensure repeatable builds
* Probably some other stuff, too; I haven't looked too closely at all the files.

cmeeren:
Partially fixes #49 

- Remove Fake and Paket (use NuGet and normal dotnet build)
- Add Directory.Build.props
- Use DotNet.ReproducibleBuilds
- Add solution files for easy access in VS/Rider
- Use GitHub Actions for CI
- Add weekly Dependabot scan
- Misc. other housekeeping

**Important:**
* If you want to always keep SDK and tool package versions in sync, the version should be moved from the `.fsproj` files to `Directory.Build.props`. Let me know and I'll move it.
* The GitHub actions NuGet push step runs only on tags starting with `v/`. In other words, when wanting to release, simply push a tag like `v/1.2.3`, and it'll be published. If you want, this can be changed, e.g. so that it always releases when the `.fsproj` version is bumped (but that makes it easier to inadvertently publish unfinished releases, so I suggest sticking to tags, which I do for all my projects.)
* The GitHub actions NuGet push step depends on a repo secret called `NUGET_API_KEY`. Let me know if you are unsure how to  set this, and I can guide you. (I cannot do it for you since it requires contributor/owner access to the repo as well as the NuGet package.)
* ~You may have to actually "activate" GitHub actions once (and point it to the file in the repo) before the pipeline starts working.~ Edit: It seems this isn't needed.
* There are no automatic GitHub releases (like [this](https://github.com/ionide/FSharp.Analyzers.SDK/releases/tag/0.11.0)). Personally I don't see the value in them for .NET OSS projects, since between published NuGet packages and a single `RELEASE_NOTES.md` package consumers have everything they need. If you really want them, it's very easy to create them manually - point it to the relevant tag and copy the release notes from `RELEASE_NOTES.md` to the GH release description. There is probably a way to add them automatically, too, but I don't know anything about that.
* `global.json` requires the exact version specified (no roll-forward). Note that this applies to development of the code in this repo, not on the runtime behavior of the tool (which is still `LatestMajor`). The reason is to avoid builds breaking because of bugs like https://github.com/dotnet/fsharp/issues/13678. We should strive to always keep the `global.json` version up to date with the latest patch version of the .NET SDK, since AFAIK it's not possible (or at least not easy) to install a lower patch (but same major/minor) version of a .NET SDK than what you have already installed. In any case, that's a change that can be done simply by updating the version number in `global.json`; no other work is necessary. If CI passes (which it will if there are no upstream bugs like the one linked to above), it's fine.
* Dependabot runs weekly and scans for outdated packages. **These serve as notifications. Do not merge the Dependabot PRs.** Dependabot does not yet handle NuGet lock files, so its PRs only update the references in the `.fsproj` files. This breaks the build. Always update the `.fsproj` packages in an IDE (or run `dotnet restore --force-evaluate` after changing the files) so that the lock files are updated.

Note that I had to add a reference to Microsoft.Build.Framework to the CLI project in order to get project loading to work.

Next up after merging this PR is more housekeeping, including Fantomas.


.idea/.idea.FSharp.Analyzers.SDK/.idea/.gitignore
```diff
@@ -0,0 +1,13 @@
+﻿# Default ignored files
```
baronfel, line 1: Can the `.idea` files/folders be removed from this PR?

Directory.Build.props
```diff
@@ -0,0 +1,15 @@
+<Project>
+
+  <Sdk Name="DotNet.ReproducibleBuilds.Isolated" Version="1.1.1" />
+
+  <PropertyGroup>
+    <TargetFramework>net6.0</TargetFramework>
```
baronfel, line 6: Please leave TFMs in individual projects - there are subtleties to project files that various tooling (not just Ionide, but VS/Rider occasionally as well) get hung up on when TFM isn't in the 'raw' project.

global.json
```diff
@@ -1,6 +1,6 @@
 {
-    "sdk": {
-        "version": "6.0.200",
-        "rollForward": "latestMajor"
-    }
+  "sdk": {
+    "version": "6.0.404",
```
baronfel, line 3: Feel free to bump to 7 SDKs.

global.json
```diff
@@ -1,6 +1,6 @@
...
+    "rollForward": "disable"
```
baronfel, line 4: unless there's a _really_ compelling, specific reason to stay on 6.0.404, I would want this line removed.

src/FSharp.Analyzers.Cli/FSharp.Analyzers.Cli.fsproj
```diff
@@ -1,20 +1,34 @@
...
+    <RollForward>LatestMajor</RollForward>
     <OutputType>Exe</OutputType>
-    <TargetFramework>net6.0</TargetFramework>
+    <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
     <PackAsTool>true</PackAsTool>
     <PackageId>fsharp-analyzers</PackageId>
     <ToolCommandName>fsharp-analyzers</ToolCommandName>
-    <RollForward>LatestMajor</RollForward>
+    <Version>0.11.0</Version>
+    <PackageReleaseNotes>https://github.com/ionide/FSharp.Analyzers.SDK/blob/master/RELEASE_NOTES.md</PackageReleaseNotes>
```
baronfel, line 11: Putting a URL here has a bad experience on NuGet.org - instead please use something like [this MSBuild target](https://github.com/dotnet/sdk-container-builds/blob/main/packaging/package.csproj#L37-L42) to read the RELEASE_NOTES.md file and put its contents directly in the release notes of the package.  This will let people view the release notes entirely from NuGet.org.

This is one thing that Ionide.KeepAChangelog does really nicely (centralize the versioning/release notes part of releasing) and it's something I want in every package I maintain.

src/FSharp.Analyzers.SDK/FSharp.Analyzers.SDK.fsproj
```diff
@@ -1,25 +1,26 @@
...
-    <PackageDescription>SDK for building custom analyzers for FSAC / F# editors</PackageDescription>
+    <Description>SDK for building custom analyzers for FSAC / F# editors</Description>
     <Copyright>Copyright 2019 Lambda Factory</Copyright>
     <PackageLicenseExpression>MIT</PackageLicenseExpression>
     <PackageTags>F#;analyzers;compiler;tooling;editor;</PackageTags>
-    <RepositoryUrl>https://github.com/ionide/FSharp.Analyzers.SDK</RepositoryUrl>
-    <RepositoryType>git</RepositoryType>
-    <RepositoryBranch>master</RepositoryBranch>
-    <DebugType>embedded</DebugType>
+    <PackageReleaseNotes>https://github.com/ionide/FSharp.Analyzers.SDK/blob/master/RELEASE_NOTES.md</PackageReleaseNotes>
```
baronfel, line 12: Same comment about release notes here - the target could even be shared and placed in `Directory.Build.targets`.

baronfel (commented):
Some notes while looking over this with coffee this morning. Great work so far!

baronfel:
RE automatic releases, this is one reason I made Ionide.KeepAChangelog - I can track everything in Changelog.md, seamlessly have that data embedded in my generated packages, and also have a format that has easy integrations with tools from other ecosystems (e.g. Github actions CI releases: https://github.com/fsharp/FsAutoComplete/blob/main/.github/workflows/release.yml#L34-L46)

global.json
cmeeren: I detailed this in the PR description. Did you read that?

baronfel: I did - I just vehemently disagree with the position (as the .NET SDK PM :P)

.idea/.idea.FSharp.Analyzers.SDK/.idea/.gitignore
cmeeren: Done.

global.json
cmeeren: DOne.

global.json
cmeeren: I had no idea you were the .NET SDK PM 🙂

Not that it changes my argument, but it's a great learning opportunity for me. So let me be more specific so that you can enlighten me as to why I am probably wrong. (And I can of course remove this line even if we agree to disagree.)

In https://github.com/dotnet/fsharp/issues/13678, the OP experienced builds sudenly breaking on CI. I experienced the same at my workplace, which caused some consternation as it broke all our CI builds and prevented me from publishing updates to our services for some hours while I debugged the issue and fixed the build.

The issue only happens if you follow certain best practices regarding repeatable builds, such as restoring packages with lock files (and maybe other stuff, e.g. added by Dotnet.ReproducibleBuilds). The core of the issue was that the FSharp.Core package shipped with the SDK and the corresponding version of FSharp.Core on NuGet had different content hashes. (As I understand it, this was an error on the F# team's part, the details of which can be read in the thread.) The result was that you could install/restore FSharp.Core locally, but not on build servers, depending on the environment on the dev machines and servers (the .NET SDK version, I guess).

One very simple workaround (that saved my day) was to lock the .NET SDK to the previous (patch) version, which did not have this issue. Hopefully, the issue would then be fixed in the next SDK update. (The issue actually occurred again just after, before they finally fixed it.)

Having been burned by this, I chose to pin the exact SDK version at work, and later also for my OSS projects. My experience was that disabling roll-forward was necessary for repeatable builds, since it broke and caused trouble for us when I didn't. You live, you learn. According to the research I did, pinning the .NET SDK version is something that most people don't do - it's a tradeoff of up-to-date SDKs vs. completely repeatable builds - but since I was actually burned by it, due to factors outside my control, I chose to do it for my projects.

I'm sure there are errors in my reasoning, and I look forward to being corrected and learning more. 🙂 For example, the issue discusses `DisableImplicitLibraryPacksFolder`, which (despite me having participated actively in that discussion) I still  don't fully understand. Reading over it again now, it seems like enabling this would mean that `FSharp.Core` is always restored from NuGet, never the SDK, and therefore fix similar problems without having to pin the SDK version. Is that correct? Since `DisableImplicitLibraryPacksFolder` is set to `true` in this PR, that would mean that pinning (here and in all my other projects, both OSS and at work) is unnecessary.

By the way, why specify a verion in `global.json` at all - or at least, why specify .NET 7, as you requested above? Why not specify the lowest version that will work in `global.json`, and allow major version roll forward?

baronfel: This is an excellent comment, but I will be away for the next couple hours on errands so it'll be a bit before I can give it the response it deserves 😅

Directory.Build.props
cmeeren: I see. I have never experienced this in Rider. Could you give me some more information, or links where I can read more?

I'll have this fixed shortly.

src/FSharp.Analyzers.Cli/FSharp.Analyzers.Cli.fsproj
cmeeren: I think Ionide.KeepAChangelog looks like a great tool. Do you want me to add that instead of the target you link to? And would Ionide.KeepAChangelog require renaming `RELEASE_NOTES.md` to `CHANGELOG.md`?

cmeeren:
> RE automatic releases [...] (e.g. Github actions CI releases: https://github.com/fsharp/FsAutoComplete/blob/main/.github/workflows/release.yml#L34-L46)

If you want to modify the release steps (currently only NuGet push) and/or trigger (currently a tag that starts with `v/`), let me know if I can help. Alternatively, you can do it after this PR is merged. It's all the same to me, as long as I know what to add where. (You can also do it yourself in this PR, since maintainer edits are allowed.)

global.json
baronfel: The first major point you seem to be making is around point-in-time fixes for specific issues. In that one, singular, circumstance it's fine to pin explicitly with a combination of version number + disabling roll forward.  In general, though, we hope to provide ways to workaround specific problems (like the FSharp.Core one you link to here) so that the uninterrupted update of SDK tooling can continue.  For this specific case, the workaround that would allow continued updates would be to disable the implicit FSharp.Core reference and pin to a precise one from NuGet, along with disabling the SDK-default 'implicit library packs' mechanism.

The implicit library packs mechanism is a way for the SDK to ship specific versions of packages as part of the SDK install. The main goal of this is to allow more 'zero restore required' workflows - i.e. you should be able to create and build an F# project without needing to connect to NuGet.org.  This however relies on the packages being shipped in the SDK to match exactly with this from NuGet (as you found).  

Overall, if you disable this mechanism entirely, then you should be able to manage FSharp.Core manually and have it always be restored from NuGet rather than the SDK.

Going forward, the SDK is planning to work with the NuGet team to make the 'library packs' mechanism work with features like Package Source Mapping as well, there's a bit of friction between the mechanisms currently.

src/FSharp.Analyzers.Cli/FSharp.Analyzers.Cli.fsproj
baronfel: Moving to Ionide.KeepAChangelog would be great, but it comes with two caveats:

* the existing README.md file would need to be converted to the Conventional Changelog format in its entirety. This is a bit of a pain, so I made a FAKE task to help with it when I [converted FsAutoComplete](https://github.com/fsharp/FsAutoComplete/pull/881/commits/af8118729e9737071c543de58223164134051fe4#diff-5ebd60ceb88910043c04c5a665d836197d7184488edcf52f590e8fcad6101519R196-R208)
* Ionide.KeepAChangelog has a feature gap where it will not load in Visual Studio at all (because VS using .NET framework MSBuild and has assembly load errors I haven't figured out yet).  Moving to it would mean we'd either forbid VS users from contributing to this project, or (more likely) we'd need a `Condition` on the PackageReference to only add it if `'$(BuildingInsideVisualStudio)' != 'true'`

global.json
cmeeren: I see. I have now removed that line. However, I have several questions:
* This defaults `rollForward` to `latestPatch`. If we always want to use the latest .NET SDK, should we set `latestMajor`? If not, why not, and then what about other `rollForward` values?
* If we want `latestMajor`, then why have a `global.json` at all? As I understand it, not having it is the same as `rollForward` set to `latestMajor`, except that when it's present, we require at least the version specified in `global.json`. Unless we need an SDK version that is so recent that we can't expect contributors or build servers to have it, it seems we can do without `global.json`.
* If we want `latestMajor` and despite my reasoning above still want `global.json`, why set it to a recent SDK version and not the earliest possible version that will work?

src/FSharp.Analyzers.Cli/FSharp.Analyzers.Cli.fsproj
cmeeren: It works fine in VS 17.4.2. I will convert the changelog and add Ionide.KeepAChangelog to this PR.

cmeeren: Note that I assume you meant converting the changelog (not the readme) to the "Keep a changelog" format (not "Common  changelog", which seems to be [something else](https://github.com/conventional-changelog/conventional-changelog)).

cmeeren:
Do you want to always keep SDK and tool package versions in sync? If so, I can move the version from the `.fsproj` files to `Directory.Build.props`.

global.json
baronfel: There are I think three main use cases that global.json helps with (and that we (SDK team) should support and document):

1. I want to be on latest all the time - aka delete global.json. This is our stated goal, and anything that prevents users from doing this is A Problem to me.
2. I want to establish a minimum bar for contribution to this repository - aka global.json with a version and rollForward set to `latestMajor`. This is the next-most ideal state IMO.
3. I have a very particular bug/regression that the SDK is causing - aka global.json with version and rollForward set to `disabled`. This is the least ideal state, and should only be a short-term decision, until the next monthly SDK release.

Ideally there would be some way to express case 2 easily, because it does have value - it clearly declares what this repo needs in order to build.

baronfel:
This looks great @cmeeren - thanks for pushing on this.

global.json
cmeeren: I see. Thank you for those clarifications. I am confused, however, since you merged this PR in its current state. I would assume that you either wanted to specify `latestMajor` in `global.json`, or else remove `global.json` entirely. I'm happy to make a PR with one of those changes. Would you like that?

cmeeren: Just thought of something: Based on the [Releases and support for .NET](https://learn.microsoft.com/en-us/dotnet/core/releases-and-support) page, there can be breaking changes between major and minor SDK versions. Is it then really a good idea to allow automatic roll-forward using `latestMajor` or `latestMinor`? Couldn't that cause problems when breaking changes occur? Also (asking purely for my own interest now), does the use of self-contained apps impact that decision, since AFAIK then the runtime .NET version is determined by the SDK it's built with?

Issue #66, opened 2023-03-24
Directions in readme don't seem to work


bigjonroberts:
**Describe the bug**

**To Reproduce**
Steps to reproduce the behaviour:
Follow these directions here: https://github.com/ionide/FSharp.Analyzers.SDK#how-to-run-sample

**Expected behaviour**
The sample analyzer runs

**Screenshots**
![image](https://user-images.githubusercontent.com/3883814/227531351-b5802bed-ed66-4532-b41a-6f9b11ea6cd2.png)

**Environment (please complete the following information):**
 - OS: Windows 10
 - Ionide version: N/A
 - VSCode version: N/A
 - dotnet SDK version: 7.0.102
 - mono / .Net Framework version: N/A

dawedawe:
@bigjonroberts Could you please check if this is still the case for you with the latest README, please.

bigjonroberts:
I'll give it a try this weekend.

Numpsy:
Are there still any issues with this?

PR #69, opened 2023-05-12
Upgrade to FSharp.Core 7+ and newer FSharp.CompilerServices

DamianReeves:
This fixes issues where even the Options example in this repo doesn't work if you are using dotnet7


bigjonroberts:
Nice

bigjonroberts:
I just tried this out on Windows 10 with dotnet sdk version 7.0.101 and had a similar error as #66 .
I'll try it out under debug later this weekend and see if I can get the detailed error message.
The Analyzers need to output more detailed error messages to allow easier diagnosis of issues like this.  I might put in a PR for that ...

image

DamianReeves:
> I just tried this out on Windows 10 with dotnet sdk version 7.0.101 and had a similar error as #66 . I'll try it out under debug later this weekend and see if I can get the detailed error message. The Analyzers need to output more detailed error messages to allow easier diagnosis of issues like this. I might put in a PR for that ...
> 
> image

I think I had to do a `-c Release` when I ran.

baronfel:
Sorry for the delay here - thanks for tackling this @DamianReeves!

PR #72, opened 2023-09-04
Verify sdk version

nojaf:
It can be very confusing when your analyzer is targeting the wrong SDK and gets loaded anyway by the cli tool. I propose to enforce using the correct version instead.


TheAngryByrd (approved):
Looks great, thanks!

PR #73, opened 2023-09-05
Add symbol uses to Context

dawedawe:
To make a lot of interesting analyzers possible, we need a bit more information in the context.
So let's add the symbol uses to it.

### DIFF
--- a/src/FSharp.Analyzers.Cli/Program.fs
+++ b/src/FSharp.Analyzers.Cli/Program.fs
@@ -98,15 +98,18 @@ let getAllEntities (checkResults: FSharpCheckFileResults) (publicOnly: bool) : A
     with
     | _ -> []
 
-let createContext (allSymbolUses: FSharpSymbolUse array) (file, text: string, p: FSharpParseFileResults, c: FSharpCheckFileResults) =
+let createContext
+    (checkProjectResults: FSharpCheckProjectResults, allSymbolUses: FSharpSymbolUse array)
+    (file, text: string, p: FSharpParseFileResults, c: FSharpCheckFileResults) =
     match c.ImplementationFile with
     | Some tast ->
         let context : Context = {
+            ParseFileResults = p
+            CheckFileResults = c
+            CheckProjectResults = checkProjectResults 
             FileName = file
             Content = text.Split([|'\n'|])
-            ParseTree = p.ParseTree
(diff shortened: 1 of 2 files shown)


TheAngryByrd (approved):
Looks good! I wonder if we should just expose the ParseFileResults, CheckFileResults, and CheckProjectResults.

src/FSharp.Analyzers.Cli/Program.fs
```diff
@@ -98,16 +98,21 @@ let getAllEntities (checkResults: FSharpCheckFileResults) (publicOnly: bool) : A
...
+    (file, text: string, p: FSharpParseFileResults, c: FSharpCheckFileResults) =
     match c.ImplementationFile with
     | Some tast ->
         let context : Context = {
+            ParseFileResults = p
+            CheckFileResults = c
+            CheckProjectResults = checkProjectResults 
             FileName = file
             Content = text.Split([|'\n'|])
-            ParseTree = p.ParseTree
```
dawedawe, line 107: removed as it can now be easily obtained from `ParseFileResults`

src/FSharp.Analyzers.Cli/Program.fs
```diff
@@ -98,16 +98,21 @@ let getAllEntities (checkResults: FSharpCheckFileResults) (publicOnly: bool) : A
...
             TypedTree = tast
-            Symbols = c.PartialAssemblySignature.Entities |> Seq.toList
```
dawedawe, line 109: removed as it can now be easily obtained from CheckFileResults

dawedawe:
> Looks good! I wonder if we should just expose the ParseFileResults, CheckFileResults, and CheckProjectResults.

Nice. I removed stuff I think is easy enough to get your hands on from the new record fields.
Other fields I kept for performance and ergonomic reasons.
Let me know, what you think.

TheAngryByrd (approved):

TheAngryByrd:
Thank you! 🍻

baronfel:
Quick proposal with zero thought:
* make the check results fields `option`s, and call the analyzer when any of the following happen:
  * file parsed
  * file checked
  * project checked
progressively filling in pieces of data as we have them.

Analyzer authors can ignore the calls with contexts that lack pieces of data that are required for them?

PR #76, opened 2023-09-07
don't process SDK dll itself

dawedawe:
Now with the filter/error reporting regarding mismatched referenced SDK versions from [72](https://github.com/ionide/FSharp.Analyzers.SDK/pull/72), loading and processing the SDK dll itself is not a good idea as the `Array.find` call in the filter throws with it.


nojaf (approved):

TheAngryByrd (approved):
Thanks!

PR #77, opened 2023-09-07
update to projinfo 0.62 to fix runtime failure

dawedawe:
We need this to let the new version of FCS and projinfo align with each other.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,3 +30,6 @@ jobs:
 
       - name: Test
         run: dotnet test -c Release --no-build
+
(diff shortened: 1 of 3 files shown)


nojaf (approved):

nojaf:
Maybe also run the sample in CI in this PR?

dawedawe:
> Maybe also run the sample in CI in this PR?

Here or in another PR. But yes, we should do it as a runtime test.
Currently it breaks because we need the fix from #76

TheAngryByrd (approved):
🎸

PR #85, opened 2023-09-11
Use build script

nojaf:
Extract all the build steps from the GitHub Actions to an F# build script.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,20 +19,5 @@ jobs:
       - name: Setup .NET
         uses: actions/setup-dotnet@v3
 
-      - name: Restore tools
-        run: dotnet tool restore
-
-      - name: Check format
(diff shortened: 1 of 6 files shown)


dawedawe (approved):

.github/workflows/ci.yml
```diff
@@ -19,20 +19,5 @@ jobs:
       - name: Setup .NET
         uses: actions/setup-dotnet@v3
 
-      - name: Restore tools
```
baronfel, line 22: As a nit, I'm less in favor of this for very simple pipelines like this - you lose the ability to see on the PR/Actions which step failed without going into the job.

nojaf: This is a trade-off scenario:
- You want to run a single command that does what CI does to verify things locally.
- You don't want that single command to be duplicated in an action's yaml as it could get out of sync.

PR #88, opened 2023-09-11
Test infrastructure

dawedawe:
This refactors some code to make unit testing with the SDK easier.
To make really sure we are using valid `FSharpProjectOptions` with all platforms and framework versions we make some effort to create a temporary project and read the options from it.
~~This is currently not working for net8, see [here](https://github.com/ionide/FSharp.Analyzers.SDK/pull/88/files#diff-60f697201de77bcf46e2e07856e6fab988f9eeade5ee215cebcbedd17b2e7568R190)~~


src/FSharp.Analyzers.SDK/FSharp.Analyzers.SDK.Client.fs
```diff
@@ -85,7 +85,7 @@ module Client =
         if Directory.Exists dir then
             let analyzerAssemblies =
                 Directory.GetFiles(dir, "*Analyzer*.dll", SearchOption.AllDirectories)
-                |> Array.filter (fun a -> not (a.EndsWith("FSharp.Analyzers.SDK.dll")))
+                |> Array.filter (fun a -> not (a.EndsWith("FSharp.Analyzers.SDK.dll") || a.EndsWith("Test.dll")))
```
nojaf, line 88: We probably want to Regex this check a bit.
If people decide to use `.*Tests.dll` instead.

src/FSharp.Analyzers.SDK/FSharp.Analyzers.SDK.fs
```diff
@@ -51,3 +53,101 @@ type Message =
...
+            projectCacheSize = 200,
+            keepAllBackgroundResolutions = true,
+            keepAssemblyContents = true,
+            documentSource = ds
+        )
+
+    [<RequireQualifiedAccess>]
+    type SourceOfSource =
+        | Path of string
+        | DiscreteSource of String
```
nojaf, line 134: Nit: uppercase `String`.

samples/OptionAnalyzer.Test/UnitTests.fs
```diff
@@ -0,0 +1,36 @@
...
+        {
+            Code = "OV001"
+            Fixes = []
+            Message = "Option.Value shouldn't be used"
+            Range = Range.mkRange "A.fs" (Position.mkPos 6 4) (Position.mkPos 6 16)
+            Severity = Severity.Warning
+            Type = "Option.Value analyzer"
+        }
+
+    let msgs = TestHelpers.getAnalyzerMsgs source
```
nojaf, line 34: Would it not make sense API-wise if we pass in the analyzer we want to test?
It is a little hidden what analyzers this is going to load.
I could imagine unit tests that start to fail because of the introduction of a new analyzer.

Something like `let ctx = TestHelpers.createContext source` to be passed to the `Analyzer`.
TestHelpers could also contain some assertion helpers.

src/FSharp.Analyzers.SDK/FSharp.Analyzers.SDK.TestHelpers.fs
```diff
@@ -2,16 +2,149 @@ module FSharp.Analyzers.SDK.TestHelpers
...
-open FSharp.Compiler.Text
 
-let getAnalyzerMsgs source =
-    let loadProject (fcs: FSharpChecker) fileName sourceText =
-        fcs.GetProjectOptionsFromScript(fileName, sourceText)
-        |> Async.RunSynchronously
-        |> fst
+type DotNetVersion =
+    | Six
+    | Seven
```
TheAngryByrd, line 14: Should probably have an escape hatch for future versions.

baronfel: Especially since the first production-ready release for .Net 8 drops in...a couple hours 😂

src/FSharp.Analyzers.Cli/Program.fs
```diff
@@ -145,6 +67,13 @@ let runProject toolsPath proj (globs: Glob list) =
     let allSymbolUses = checkProjectResults.GetAllUsesOfAllSymbols()
 
     opts.SourceFiles
+    |> Array.filter (fun file ->
+        match Path.GetExtension(file).ToLowerInvariant() with
+        | ".fsi" ->
+            printInfo $"Ignoring signature file %s{file}"
```
nojaf, line 73: Why is this again?

src/FSharp.Analyzers.SDK/FSharp.Analyzers.SDK.TestHelpers.fsi
```diff
@@ -0,0 +1,19 @@
...
+
+    override ToString: unit -> string
+
+type FSharpProjectOptions with
+
+    static member zero: FSharpProjectOptions
+
+type Package = { Name: string; Version: string }
+
+val mkOptionsFromProject: version: DotNetVersion -> additionalPkg: Package option -> FSharpProjectOptions
```
nojaf, line 17: Maybe rename `version` to `framework and pass in a string instead?
Having a proper XML comment can point out what we are going to do with it.
`additionalPkg` should be a list.
The whole thing should be `Async<>` or `Task<>`

src/FSharp.Analyzers.SDK/FSharp.Analyzers.SDK.TestHelpers.fs
```diff
@@ -0,0 +1,187 @@
...
+        UnresolvedReferences = None
+        OriginalLoadReferences = []
+        Stamp = None
+    }
+
+let mkOptionsFromBinaryLog binLogPath =
+    let compilerArgs = readCompilerArgsFromBinLog binLogPath
+    mkOptions compilerArgs
+
+let mkOptionsFromProject (version: DotNetVersion) (additionalPkg: Package option) =
```
nojaf, line 109: This needs a `try/with`, so much stuff that could go wrong.
I also think we need to reconsider how we make the binlog file.
If I run my tests twice I don't want to have two binlog files that carry the same information.
The second run should reuse the evaluation of the first run.

Proposal: Create a unique file name based on the input parameters. Inside the temp folder, we create an `FSharp.Analyzers.SDK` cache folder (where we store all the bin logs).
If the file doesn't exist yet, we do the `dotnet new classlib` dance, otherwise, we read the binlog again. The SDK version should also be included in the unique file name.

For example, something along the lines of:
`mkOptionsFromProject "net6.0" [ { Name = "Fantomas.Core"; Version = "6.2" } ]` would be
`v0.13.0_net6.0_Fantomas.Core_6.2`

src/FSharp.Analyzers.Cli/Program.fs
dawedawe: bad merge, sorry.

src/FSharp.Analyzers.SDK/FSharp.Analyzers.SDK.TestHelpers.fsi
```diff
@@ -0,0 +1,30 @@
...
+/// <param name="framework">The target framework for the tested code to use. E.g. net6.0, net7.0</param>
+/// <param name="additionalPkgs">A list of additional packages that should be referenced. The tested code can use these.</param>
+/// <returns>FSharpProjectOptions</returns>
+val mkOptionsFromProject: framework: string -> additionalPkgs: Package list -> Task<FSharpProjectOptions>
+
+val getContext: opts: FSharpProjectOptions -> source: string -> Context
+
+module AssertionHelpers =
+
+    val areWarningsInLines: msgs: Message list -> expectedLines: Set<int> -> bool
```
nojaf, line 22: Nitpick: should the `expectedLines` not come first? And maybe rename it to `hasWarningsinLines`?

src/FSharp.Analyzers.SDK/FSharp.Analyzers.SDK.TestHelpers.fsi
```diff
@@ -0,0 +1,30 @@
+module FSharp.Analyzers.SDK.TestHelpers
```
nojaf, line 1: Nitpick: I'm on the fence about words like `Helpers`.
Maybe go with `Testing` and later on with `Assert`

nojaf (approved):
Some nits left but I'm really happy with this setup.
Please consider adding some documentation about this as well.
Something a la "Write your first unit test"

nojaf (approved):
Great work! I'm going to merge this in to play around with the testing story in my current work.

PR #90, opened 2023-09-13
Revisit Context type

nojaf:
A slight restructure of the `Context` type.
I think for now it should really expose all the base FCS primitives so people can do everything they need.

I've added some convenient members and we can expand upon that later.
I'm not quite sure yet about `CheckProjectResults: Async<FSharpCheckProjectResults>` yet.
As in should it be wrapped in `lazy` as well?

I quite like that `GetAllEntities`, `AllSymbolUses` and `SymbolUsesOfFile` are more on a demand basis.
Not all analyzers will require this.

Let me know what you think.

### DIFF
--- a/docs/content/Dual Analyzer.fsx
+++ b/docs/content/Dual Analyzer.fsx
@@ -17,6 +17,7 @@ With a little orchestration it is possible to easily write two analyzer function
 (** *)
 
 open FSharp.Analyzers.SDK
+open FSharp.Compiler.CodeAnalysis
 open FSharp.Compiler.Text
 open FSharp.Compiler.Syntax
 
@@ -24,23 +25,30 @@ open FSharp.Compiler.Syntax
 /// See https://learn.microsoft.com/en-us/dotnet/fsharp/style-guide/conventions#sort-open-statements-topologically
 /// Note that this implementation is not complete and only serves as an illustration.
 /// Nested modules are not taken into account.
-let private topologicallySortedOpenStatementsAnalyzer (untypedTree: ParsedInput) : Async<Message list> =
+let private topologicallySortedOpenStatementsAnalyzer
+    (sourceText: ISourceText)
+    (untypedTree: ParsedInput)
+    (checkResults: FSharpCheckFileResults)
+    : Async<Message list>
+    =
     async {
         let allOpenStatements =
-            let allOpenStatements = ResizeArray<string * range>()
+            let allOpenStatements = ResizeArray<string list * range>()
 
             let (|LongIdentAsString|) (lid: SynLongIdent) =
-                lid.LongIdent |> List.map (fun ident -> ident.idText) |> String.concat "."
+                lid.LongIdent |> List.map (fun ident -> ident.idText)
 
             let rec visitSynModuleSigDecl (decl: SynModuleSigDecl) =
                 match decl with
-                | SynModuleSigDecl.Open(SynOpenDeclTarget.ModuleOrNamespace(longId = LongIdentAsString value), mOpen) ->
+                | SynModuleSigDecl.Open(
+                    target = SynOpenDeclTarget.ModuleOrNamespace(longId = LongIdentAsString value; range = mOpen)) ->
                     allOpenStatements.Add(value, mOpen)
                 | _ -> ()
 
             let rec visitSynModuleDecl (decl: SynModuleDecl) =
                 match decl with
-                | SynModuleDecl.Open(SynOpenDeclTarget.ModuleOrNamespace(longId = LongIdentAsString value), mOpen) ->
+                | SynModuleDecl.Open(
+                    target = SynOpenDeclTarget.ModuleOrNamespace(longId = LongIdentAsString value; range = mOpen)) ->
                     allOpenStatements.Add(value, mOpen)
                 | _ -> ()
 
@@ -57,17 +65,32 @@ let private topologicallySortedOpenStatementsAnalyzer (untypedTree: ParsedInput)
 
             allOpenStatements |> Seq.toList
 
-        let isOpenStatement (openStatement: string, _) = openStatement.StartsWith("System")
+        let isSystemOpenStatement (openStatement: string list, mOpen: range) =
+            let isFromBCL () =
+                let line = sourceText.GetLineString(mOpen.EndLine - 1)
 
-        let nonSystemOpens = allOpenStatements |> List.skipWhile isOpenStatement
+                match checkResults.GetSymbolUseAtLocation(mOpen.EndLine, mOpen.EndColumn, line, openStatement) with
+                | Some symbolUse ->
+                    match symbolUse.Symbol.Assembly.FileName with
+                    | None -> false
+                    | Some assemblyPath ->
+                        // This might not be an airtight check
+                        assemblyPath.ToLower().Contains "microsoft.netcore.app.ref"
+                | _ -> false
+
+            openStatement.[0].StartsWith("System") && isFromBCL ()
+
+        let nonSystemOpens = allOpenStatements |> List.skipWhile isSystemOpenStatement
 
         return
             nonSystemOpens
-            |> List.filter isOpenStatement
+            |> List.filter isSystemOpenStatement
             |> List.map (fun (openStatement, mOpen) ->
+                let openStatementText = openStatement |> String.concat "."
+
                 {
                     Type = "Unsorted System open statement"
-                    Message = $"%s{openStatement} was found after non System namespaces where opened!"
+                    Message = $"%s{openStatementText} was found after non System namespaces where opened!"
                     Code = "SOT001"
                     Severity = Warning
                     Range = mOpen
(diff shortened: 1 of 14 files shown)


src/FSharp.Analyzers.SDK/FSharp.Analyzers.SDK.fs
```diff
@@ -17,20 +19,65 @@ type AnalyzerAttribute([<Optional; DefaultParameterValue "Analyzer">] name: stri
...
         ParseFileResults: FSharpParseFileResults
         CheckFileResults: FSharpCheckFileResults
-        CheckProjectResults: FSharpCheckProjectResults
-        FileName: string
-        Content: string[]
         TypedTree: FSharpImplementationFileContents
-        GetAllEntities: bool -> AssemblySymbol list
-        AllSymbolUses: FSharpSymbolUse array
-        SymbolUsesOfFile: FSharpSymbolUse array
+        CheckProjectResults: Async<FSharpCheckProjectResults>
```
TheAngryByrd, line 27: After mulling over https://github.com/ionide/FSharp.Analyzers.SDK/pull/73#issuecomment-1709336286, I think it may be better to either encode these as some form of Options or do the "tiered" approach suggested by @dawedawe (like individual parse/checkfile/checkproject events). Reason being, if this is encoded as `Async`, any analyzer can ask for this every time, and FSAC won't be in control of when to get `CheckProjectResults`. In other worse, someone is going to write an analyzer that can tank performance.

TheAngryByrd:
We talked a bit on Discord about the different contexts an analyzer can be run in, either CLI or IDE.

CLI typically has everything upfront, however the IDE has specific performance concerns such as ProjectCheckResults not always available and Cancellation of a running analyzer.

We're considering having "two" entry points for these different concerns.

1. Attribute for CLI with all the properties filled out
2. Attribute for IDE with `Optional` fields that the analyzer can then call its base implementation if it pattern matches on the fields it requires. It will also have a `CancellationToken` or be `Async`.


cc @baronfel

TheAngryByrd (approved):
Lovely! Looks good to me.  Added some small things to consider since we're revamping it.

docs/content/Dual Analyzer.fsx
```diff
@@ -0,0 +1,94 @@
...
+[<EditorAnalyzer "Topologically sorted open statements">]
+let editorAnalyzer (ctx: EditorContext) =
+    match ctx.ParseFileResults with
+    // The editor might not have any parse results for a given file. So we don't return any messages.
+    | None -> []
+    | Some parseResults -> topologicallySortedOpenStatementsAnalyzer parseResults.ParseTree
+
+(**
+Both analyzer will follow the same code path: the console application will always have the required data, while the editor needs to be more careful.  
+⚠️ Please do not be tempted by calling `.Value` on the `EditorContext` 😉.
```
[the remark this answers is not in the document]
TheAngryByrd: > ⚠️ Please do not be tempted by calling `.Value` on the `EditorContext` 😉.

If they use the option analyzer they wont be tempted to! 🙃

TheAngryByrd (approved):

nojaf:
One thing I'm still wondering about is the parse results.
If those are empty, it means the IDE has absolutely nothing right?
Will it ever invoke any analyzers without it?

docs/content/Dual Analyzer.fsx
```diff
@@ -0,0 +1,94 @@
...
+[<CliAnalyzer "Topologically sorted open statements">]
+let cliAnalyzer (ctx: CliContext) =
+    topologicallySortedOpenStatementsAnalyzer ctx.ParseFileResults.ParseTree
+
+[<EditorAnalyzer "Topologically sorted open statements">]
+let editorAnalyzer (ctx: EditorContext) =
+    match ctx.ParseFileResults with
+    // The editor might not have any parse results for a given file. So we don't return any messages.
+    | None -> []
+    | Some parseResults -> topologicallySortedOpenStatementsAnalyzer parseResults.ParseTree
```
TheAngryByrd, line 85: Hmm, is there a way to make this async or take a `Cancellationtoken`? I think it would be useful to have cancellation as a possibility for editor scenarios. (I know people might not respect them but that's another concern)

TheAngryByrd:
> One thing I'm still wondering about is the parse results. If those are empty, it means the IDE has absolutely nothing right? Will it ever invoke any analyzers without it?

I can't think of a reason we would want to run without that info.

nojaf:
> I can't think of a reason we would want to run without that info.

Do we want to lift the option for the parse results then?

TheAngryByrd:
> Do we want to lift the option for the parse results then?

I think it's fine to always have it. Might hurt your example though :)

nojaf:
> Might hurt your example though :)

Yup, I'll need to cook up something else for that.

TheAngryByrd:
> Yup, I'll need to cook up something else for that.

Could show off the Option analyzer as it's already the hello world analyzer we have.

TheAngryByrd (approved):
Ok I think all my concerns are addressed.  LGTM.

src/FSharp.Analyzers.SDK/FSharp.Analyzers.SDK.fsi
```diff
@@ -0,0 +1,121 @@
...
+        /// Represents the results of parsing an F# file and a set of analysis operations based on the parse tree alone.
+        /// See <a href="https://fsharp.github.io/fsharp-compiler-docs/reference/fsharp-compiler-codeanalysis-fsharpparsefileresults.html">FSharpParseFileResults Type</a>
+        ParseFileResults: FSharpParseFileResults
+        /// A handle to the results of CheckFileInProject.
+        /// See <a href="https://fsharp.github.io/fsharp-compiler-docs/reference/fsharp-compiler-codeanalysis-fsharpcheckfileresults.html">FSharpCheckFileResults Type</a>
+        CheckFileResults: FSharpCheckFileResults
+        /// Represents the definitional contents of a single file or fragment in an assembly, as seen by the F# language
+        /// See <a href="https://fsharp.github.io/fsharp-compiler-docs/reference/fsharp-compiler-symbols-fsharpimplementationfilecontents.html">FSharpImplementationFileContents Type</a>
+        TypedTree: FSharpImplementationFileContents
+        /// A handle the results of the entire project
```
dawedawe, line 52:
```suggestion
        /// A handle to the results of the entire project
```

src/FSharp.Analyzers.SDK/FSharp.Analyzers.SDK.fsi
```diff
@@ -0,0 +1,121 @@
...
+        /// Represents the results of parsing an F# file and a set of analysis operations based on the parse tree alone.
+        /// See <a href="https://fsharp.github.io/fsharp-compiler-docs/reference/fsharp-compiler-codeanalysis-fsharpparsefileresults.html">FSharpParseFileResults Type</a>
+        ParseFileResults: FSharpParseFileResults option
+        /// A handle to the results of CheckFileInProject.
+        /// See <a href="https://fsharp.github.io/fsharp-compiler-docs/reference/fsharp-compiler-codeanalysis-fsharpcheckfileresults.html">FSharpCheckFileResults Type</a>
+        CheckFileResults: FSharpCheckFileResults option
+        /// Represents the definitional contents of a single file or fragment in an assembly, as seen by the F# language
+        /// See <a href="https://fsharp.github.io/fsharp-compiler-docs/reference/fsharp-compiler-symbols-fsharpimplementationfilecontents.html">FSharpImplementationFileContents Type</a>
+        TypedTree: FSharpImplementationFileContents option
+        /// A handle the results of the entire project
```
dawedawe, line 84:
```suggestion
        /// A handle to the results of the entire project
```

src/FSharp.Analyzers.SDK/FSharp.Analyzers.SDK.Client.fsi
```diff
@@ -0,0 +1,22 @@
...
+    /// Loads into private state any analyzers defined in any assembly
+    /// matching `*Analyzer*.dll` in given directory (and any subdirectories)
+    /// </summary>
+    /// <returns>number of found dlls matching `*Analyzer*.dll` and number of registered analyzers</returns>
+    member LoadAnalyzers: printError: (string -> unit) -> dir: string -> int * int
+    /// <summary>Runs all registered analyzers for given context (file).</summary>
+    /// <returns>list of messages. Ignores errors from the analyzers</returns>
+    member RunAnalyzers: ctx: 'TContext -> Async<Message list>
+    /// <summary>Runs all registered analyzers for given context (file).</summary>
+    /// <returns>list of results per analyzer which can either messages or an exception.</returns>
```
dawedawe, line 21:
```suggestion
    /// <returns>list of results per analyzer which can either be messages or an exception.</returns>
```

dawedawe (approved):
Some wording nitpicks from my end.

nojaf:
Thanks all!

PR #97, opened 2023-09-20
Extract any duplicatish code from the main README.

nojaf:
All code samples should really be ported to a `.fsx` file in the docs, that way they are easier to stay up to date. And I also want to avoid duplicating the docs.

### DIFF
--- a/docs/content/Getting Started.fsx
+++ b/docs/content/Getting Started.fsx
@@ -113,7 +113,7 @@ Simply run `dotnet pack --configuration Release` against the analyzer project to
 dotnet nuget push {NugetPackageFullPath} -s nuget.org -k {NugetApiKey}
 ```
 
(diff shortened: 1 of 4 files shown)


docs/content/Getting Started.fsx
```diff
@@ -89,6 +104,86 @@ dotnet tool install --global fsharp-analyzers
...
+### Packaging and Distribution
+
+Since analyzers are just .NET core libraries, you can distribute them to the nuget registry just like you would with a normal .NET package.  
+Simply run `dotnet pack --configuration Release` against the analyzer project to get a nuget package and publish it with
+
+```shell
+dotnet nuget push {NugetPackageFullPath} -s nuget.org -k {NugetApiKey}
+```
+
+However, the story is different and slightly more complicated when your analyzer package has third-party dependencies also coming from nuget. Since the SDK dynamically loads the package assemblies (`.dll` files), the assemblies of the dependencies has be there *next* to the main assembly of the analyzer. Using `dotnet pack` will **not** include these dependencies into the output Nuget package. More specifically, the `./lib/net6.0` directory of the nuget package must have all the required assemblies, also those from third-party packages. In order to package the analyzer properly with all the assemblies, you need to take the output you get from running:
```
dawedawe, line 116:
```suggestion
However, the story is different and slightly more complicated when your analyzer package has third-party dependencies also coming from nuget. Since the SDK dynamically loads the package assemblies (`.dll` files), the assemblies of the dependencies have to be right *next* to the main assembly of the analyzer. Using `dotnet pack` will **not** include these dependencies into the output Nuget package. More specifically, the `./lib/net6.0` directory of the nuget package must have all the required assemblies, also those from third-party packages. In order to package the analyzer properly with all the assemblies, you need to take the output you get from running:
```

dawedawe (approved):

PR #107, opened 2023-09-26
Remove packages.lock.json files till we have a fixed sdk

dawedawe:
see https://github.com/dotnet/fsharp/issues/16014
and https://github.com/dotnet/fsharp/pull/16015


nojaf:
Sorry, was too quick here:

https://github.com/ionide/FSharp.Analyzers.SDK/blob/9df80cf2cda50b15d23826481aff0db726f80b16/Directory.Build.props#L6

Needs to be false I think.

PR #116, opened 2023-10-10
support multiple project parameters in the Cli tool

### LINKED ISSUE #113: Support multiple project parameters in the Cli tool
**Is your feature request related to a problem? Please describe.**
Calling the Cli tool multiple times for a collection of projects (e.g. in a build pipeline for a solution) wastes some CPU cycles loading the same analyzers again and again.

**Describe the solution you'd like**
I suggest we support taking in multiple projects as Cli parameters, so analyzer DLLs are only loaded once and reused for all given projects.

**Describe alternatives you've considered**
Making DLL loading and reflection so fast, that it doesn't matter. But that's above my paygrade.

**Additional context**
As always, I don't expect other people to do the work, I raise the issue to keep track of it for myself.

dawedawe:
fixes #113

### DIFF
--- a/src/FSharp.Analyzers.Cli/Program.fs
+++ b/src/FSharp.Analyzers.Cli/Program.fs
@@ -131,21 +131,17 @@ let printMessages failOnWarnings (msgs: Message list) =
 
     msgs
 
-let calculateExitCode failOnWarnings (msgs: Message list option array) : int =
+let calculateExitCode failOnWarnings (msgs: Message list array) : int =
     match msgs with
     | [||] -> -1
     | msgsLists ->
         let check =
             let checkList msgs =
                 msgs
-                |> Option.map (fun msgs ->
-                    msgs
-                    |> List.exists (fun n ->
-                        n.Severity = Error
-                        || (n.Severity = Warning && failOnWarnings |> List.contains n.Code)
-                    )
+                |> List.exists (fun n ->
+                    n.Severity = Error
+                    || (n.Severity = Warning && failOnWarnings |> List.contains n.Code)
                 )
-                |> Option.defaultValue false
 
             msgsLists |> Seq.exists checkList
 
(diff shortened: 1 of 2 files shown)


src/FSharp.Analyzers.Cli/Program.fs
```diff
@@ -175,27 +182,28 @@ let main argv =
...
+                        if Path.IsPathRooted proj then
+                            proj
+                        else
+                            Path.GetFullPath(Path.Combine(Environment.CurrentDirectory, proj))
+
+                    let! results = runProject client toolsPath project ignoreFiles
+                    return results |> Option.map (printMessages failOnWarnings)
+                }
+
+            projects |> List.map runProj |> Async.Sequential |> Async.RunSynchronously
```
nojaf, line 207: Any reason we don't want to run these in parallel?

src/FSharp.Analyzers.Cli/Program.fs
```diff
@@ -131,16 +131,23 @@ let printMessages failOnWarnings (msgs: Message list) =
 
     msgs
 
-let calculateExitCode failOnWarnings (msgs: Message list option) : int =
+let calculateExitCode failOnWarnings (msgs: Message list option array) : int =
```
nojaf, line 134: Can we not flatten this collection earlier? This type of definition is getting out of hand.

nojaf (approved):
Looks good, I'm wondering a bit about the parallel stuff.

src/FSharp.Analyzers.Cli/Program.fs
dawedawe: Yes, I'm worried about mangled console output.

src/FSharp.Analyzers.Cli/Program.fs
dawedawe: Absolutely, good call

src/FSharp.Analyzers.Cli/Program.fs
```diff
@@ -131,10 +131,10 @@ let printMessages failOnWarnings (msgs: Message list) =
 
     msgs
 
-let calculateExitCode failOnWarnings (msgs: Message list option) : int =
+let calculateExitCode failOnWarnings (msgs: Message list) : int =
     match msgs with
-    | None -> -1
-    | Some msgs ->
+    | [] -> -1
```
nojaf, line 136: This doesn't seem right. Exitcode should be 0 if no messages are found I'd say.

PR #117, opened 2023-10-12
Throw custom exception with diagnostics if the severity is Error

dawedawe:
Why should we do this?
It's very easy to construct tests which aren't valid F# but still produce tree structures that are passed on via the context. 
That could erroneously be processed by analyzers and give wrong test results.

Another pitfall is to use an API in the testcase, which isn't available in netstandard.
If the analyzer author writes a test for that, his analyzer could process the information from the context even though it wouldn't compile in the real world due to a missing overload in the netstandard API.

So, to make the analyzer/test author aware of errors in his testcases, we throw an exception with diagnostics of error severity.

### DIFF
--- a/src/FSharp.Analyzers.SDK.Testing/FSharp.Analyzers.SDK.Testing.fsi
+++ b/src/FSharp.Analyzers.SDK.Testing/FSharp.Analyzers.SDK.Testing.fsi
@@ -10,7 +10,7 @@ type FSharpProjectOptions with
 
 type Package = { Name: string; Version: string }
 
-exception DiagnosticErrors of FSharpDiagnostic array
+exception CompilerDiagnosticErrors of FSharpDiagnostic array
 
 /// <summary>Creates a classlib project in a temporary folder to gather the needed FSharpProjectOptions.</summary>
 /// <param name="framework">The target framework for the tested code to use. E.g. net6.0, net7.0</param>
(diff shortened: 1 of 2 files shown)


src/FSharp.Analyzers.SDK.Testing/FSharp.Analyzers.SDK.Testing.fsi
```diff
@@ -2,13 +2,16 @@ module FSharp.Analyzers.SDK.Testing
...
 open FSharp.Compiler.CodeAnalysis
+open FSharp.Compiler.Diagnostics
 
 type FSharpProjectOptions with
 
     static member zero: FSharpProjectOptions
 
 type Package = { Name: string; Version: string }
 
+exception DiagnosticErrors of FSharpDiagnostic array
```
nojaf, line 13: Should we suffix this with `Exception`, feels like the dotnet thing to do here?

nojaf (approved):

src/FSharp.Analyzers.SDK.Testing/FSharp.Analyzers.SDK.Testing.fsi
nojaf: Hmm, nah that doesn't seem to be a convention in the fsharp compiler codebase.
We could maybe go with `CompilerDiagnosticErrors` but I'm not hellbent on that either.

Issue #118, opened 2023-10-12
Cli option --fail-on-all-warnings


dawedawe:
**Is your feature request related to a problem? Please describe.**
Currently, to fail on a warning, you have to list all the analyzer codes you want to fail on. If you are using many analyzers and you want to fail on all of them, that can be cumbersome.

**Describe the solution you'd like**
Provide a --fail-on-all-warnings Cli option.

**Describe alternatives you've considered**
Support globbing of analyzer codes.

**Additional context**
As always, I just raise this issue to keep track of it.

baronfel:
Consider a syntax like NoWarn in the compiler to allow setting diagnostic levels on a per message basis, as well as the more general 'treatwarningsaserror' mechanism: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-options/errors-warnings

I think it would be good for the analyzer SDK to onboard to these mechanisms because that's how they would be handled/integrated in the 'final' analyzer setup, when they're part of the compiler.

PR #122, opened 2023-10-21
Add 'Fail_On_All_Warnings' CLI Argument

1eyewonder:
Saw #118 and thought I'd take a stab at it. Per https://github.com/ionide/FSharp.Analyzers.SDK/issues/118#issuecomment-1759566818, I read through the documentation and implemented what I interpreted. If I misread, let me know and I can refactor accordingly.

### DIFF
--- a/src/FSharp.Analyzers.Cli/Program.fs
+++ b/src/FSharp.Analyzers.Cli/Program.fs
@@ -13,6 +13,7 @@ type Arguments =
     | Project of string list
     | Analyzers_Path of string
     | Fail_On_Warnings of string list
+    | Fail_On_All_Warnings of except: string list
     | Ignore_Files of string list
     | Exclude_Analyzer of string list
     | Report of string
@@ -25,10 +26,16 @@ type Arguments =
             | Analyzers_Path _ -> "Path to a folder where your analyzers are located."
             | Fail_On_Warnings _ ->
                 "List of analyzer codes that should trigger tool failures in the presence of warnings."
+            | Fail_On_All_Warnings _ ->
+                "All analyzer codes will trigger tool failure in the presence of warnings, except for the ones listed here."
             | Ignore_Files _ -> "Source files that shouldn't be processed."
             | Exclude_Analyzer _ -> "The names of analyzers that should not be executed."
             | Report _ -> "Write the result messages to a (sarif) report file."
             | Verbose -> "Verbose logging."
+
+type WarningConfig =
+    | FailOnWarnings of string list
+    | FailOnAllWarnings of except: string list
 
 let mutable verbose = false
 
@@ -111,7 +118,7 @@ let runProject (client: Client<CliAnalyzerAttribute, CliContext>) toolsPath proj
                 ]
     }
 
-let printMessages failOnWarnings (msgs: AnalyzerMessage list) =
+let printMessages warningConfig (msgs: AnalyzerMessage list) =
     if verbose then
         printfn ""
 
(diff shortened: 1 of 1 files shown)


nojaf:
Hi there! I appreciate your interest.
At first glance, I wonder if we need to shuffle some more.

Maybe sticking to the [compiler options](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-options/errors-warnings), makes more sense here:
```
WarningLevel / -warn: Set warning level.
AnalysisLevel: Set optional warning level.
TreatWarningsAsErrors / -warnaserror: Treat all warnings as errors
WarningsAsErrors / -warnaserror: Treat one or more warnings as errors
WarningsNotAsErrors / -warnnotaserror: Treat one or more warnings not as errors
NoWarn / -nowarn: Set a list of disabled warnings.
CodeAnalysisRuleSet / -ruleset: Specify a ruleset file that disables specific diagnostics.
ErrorLog / -errorlog: Specify a file to log all compiler and analyzer diagnostics.
ReportAnalyzer / -reportanalyzer: Report additional analyzer information, such as execution time.
```

1eyewonder:
Is the thought process behind this we have a breaking change to name in this fashion explicitly? Also, are you thinking we want to expand the scope of this issue/PR to try and tackle all the flags listed here?

nojaf:
Yes, I think a breaking change is fine as we are still in pre-release.
We broke the API a couple of times these last weeks, so I'm not overly worried about that.
In recent (private) conversations, we found the need to have more fine-grained control so revisiting makes sense.

dawedawe:
To cover all cases of bending the default [Severity](https://github.com/ionide/FSharp.Analyzers.SDK/blob/e31c6dac0c5117ba3ad9a9d989611ea02c801876/src/FSharp.Analyzers.SDK/FSharp.Analyzers.SDK.fsi#L123) level of a specific analyzer message code to another `Severity` level, we had the idea to provide the following CLI switches:
```
--TreatAsInfo MessageCode1 MessageCode2 ...
--TreatAsHint MessageCode3 MessageCode4 ...
--TreatAsWarning MessageCode5 MessageCode6 ...
--TreatAsError MessageCode7 MessageCode8 ...
```
That would make it very easy and clear for the user to state their intent and we already had a use case of users wanting to have a very fine grained control.
We could still offer CLI switches known from the compiler like `TreatWarningsAsErrors`, but internally map them to the above.
@baronfel @TheAngryByrd What do you think? Would that be okay for you?

TheAngryByrd:
Getting into the super fine control there is a few `locations` to think about.

- Per line
- Per region (a section of a file)
- Per file (we have `Ignore_Files` but that's a whole sale don't use any analyzers rather than choosing which analyzers to turn on/off per file)
- Per project

And as you've pointed out there's different ways to `treat` a `MessageCode`.

- TreatAsHint
- TreatAsInfo
- TreatAsWarning
- TreatAsError
- Disable (similar to `NoWarn` in `msbuild`, We currently have `Exclude_Analyzer` but that's for a whole analyzer and not specific code as an analyzer may have multiple `MessageCodes`).

For instance it's reasonable for someone to want to ignore `Option.value` analyzer in a specific file or even line, but not the rest of the project.

This creates a very fun matrix of support cases, some are easier than others. We only cover some of these cases in certain areas but this is the all the places someone is going to want to do some type of configuration.

baronfel:
Please rebase this against `main` to reopen the PR.

Issue #124, opened 2023-10-24
Analyzer cannot use FSharp.Core 8


nojaf:
**Describe the bug**
When your analyzer is using a higher version of F# Core (say `8.0.100-beta.23475.2`), it will fail at runtime during loading.

```
Unhandled exception. System.IO.FileNotFoundException: Could not load file or assembly 'FSharp.Core, Version=8.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'. The system cannot find the file specified.
File name: 'FSharp.Core, Version=8.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'
   at System.Signature.GetSignature(Void* pCorSig, Int32 cCorSig, RuntimeFieldHandleInternal fieldHandle, IRuntimeMethodInfo methodHandle, RuntimeType declaringType)
   at System.Reflection.RuntimeMethodInfo.<get_Signature>g__LazyCreateSignature|25_0()
   at System.Reflection.RuntimeMethodInfo.get_ReturnType()
   at FSharp.Analyzers.SDK.Client.getAnalyzerFromMemberInfo@61[TContext,a](a mi) in /_//src/FSharp.Analyzers.SDK/FSharp.Analyzers.SDK.Client.fs:line 69
   at FSharp.Analyzers.SDK.Client.analyzerFromMember[TAnalyzerAttribute,TContext](String path, MemberInfo mi) in /_//src/FSharp.Analyzers.SDK/FSharp.Analyzers.SDK.Client.fs:line 88
   at FSharp.Analyzers.SDK.Client.analyzersFromType@126.Invoke(MemberInfo mi)
   at Microsoft.FSharp.Collections.Internal.IEnumerator.choose@240.System.Collections.IEnumerator.MoveNext() in D:\a\_work\1\s\src\FSharp.Core\seq.fs:line 252
   at Microsoft.FSharp.Collections.SeqModule.ToList[T](IEnumerable`1 source) in D:\a\_work\1\s\src\FSharp.Core\seq.fs:line 1003
   at <StartupCode$FSharp-Analyzers-SDK>.$FSharp.Analyzers.SDK.Client.analyzers@200-2.Invoke(Type t)
   at Microsoft.FSharp.Collections.Internal.IEnumerator.map@128.DoMoveNext(b& curr) in D:\a\_work\1\s\src\FSharp.Core\seq.fs:line 131
[...]

PR #127, opened 2023-10-25
Exit the commandline tool if the version of FSharp.Core cannot be loaded

nojaf:
Helps in https://github.com/ionide/FSharp.Analyzers.SDK/issues/124

I think it currently is a bit of a lie that we say:
image

While FCS has:
image

I'm in favour of locking both FCS and FSharp.Core down.

### DIFF
--- a/src/FSharp.Analyzers.Cli/Program.fs
+++ b/src/FSharp.Analyzers.Cli/Program.fs
@@ -1,5 +1,6 @@
 ﻿open System
 open System.IO
+open System.Runtime.Loader
 open FSharp.Compiler.CodeAnalysis
 open FSharp.Compiler.Text
 open Argu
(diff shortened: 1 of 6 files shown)


src/FSharp.Analyzers.Cli/Program.fs
```diff
@@ -270,13 +266,28 @@ let main argv =
...
         { new Logger with
-            member _.Error msg = printError "%s" msg
+            member _.Error msg = printError msg
 
             member _.Verbose msg =
                 if verbose then
                     printInfo "%s" msg
         }
 
+    AssemblyLoadContext.Default.add_Resolving (fun ctx assemblyName ->
```
dawedawe, line 276:
```suggestion
    AssemblyLoadContext.Default.add_Resolving (fun _ctx assemblyName ->
```

dawedawe (approved):

PR #130, opened 2023-10-26
Add documentation on msbuild props when packaging.

nojaf:
Adds additional info you want to have when creating the nuget packages.


TheAngryByrd (approved):
Starting to seem like we're get into making a template territory :)

nojaf:
Yup, it is indeed going in that direction.

dawedawe (approved):

PR #131, opened 2023-10-26
Getting started for consumers/users

dawedawe:


docs/content/Getting Started.fsx
```diff
@@ -5,217 +5,124 @@ categoryindex: 1
...
+
+    </Target>
+
+</Project>
 ```
 
-### Third-party dependencies
+You may need to adjust the `Command` to be compatible with your specific analyzer. Think about how you want warnings to be treated.
+
+⚠️ Note: This uses a feature from .NET 8 to get hold of the `FscCommandLineArgs`.
```
nojaf, line 117: No, this will work for non-dotnet 8 rc 2 projects as well.
The new feature is the ability to use `--getProperty` and `--getItem` from the command line.

Here you are essentially invoking a bunch of targets to do a design time build.

I think we probably want to have a short section on project cracking:

The way the analyzers work is that we will programmatically type-check a project and process the results with our analyzers. In order to do this programmatical type-check we need to construct `FSharpProjectOptions`. This is essentially a type that represents all the fsharp compiler arguments. When using `--project`, we will use [ProjInfo](https://www.nuget.org/packages/Ionide.ProjInfo) to invoke a set of MSBuild targets in the project to perform a design-time build. A design-time build is basically an empty invocation of a build. It won't produce a binary but will have constructed the correct arguments to theoretically invoke the compiler.

When we use `--fsc-args` we do exactly the same thing, but we tell MSBuild to invoke the targets. This happens in the slew of `BeforeTargets` and the properties we specify in the `Directory.Solution.targets` (`Properties="DesignTimeBuild=True;Configuration=Release;ProvideCommandLineArgs=True;SkipCompilerExecution=True"`). It is this exact combination that will essentially do the same thing as `--project`. The interesting thing about doing it via MSBuild is that some targets will be cached and this speeds up everything.

I hope this makes a little sense.

docs/content/Getting Started.fsx
```diff
@@ -5,217 +5,124 @@ categoryindex: 1
...
 ```xml
-<PackageReference Update="FSharp.Core" Version="7.0.400" />
+<Target
+    Name="AnalyzeProject" 
+    DependsOnTargets="Restore;ResolveAssemblyReferencesDesignTime;ResolveProjectReferencesDesignTime;ResolvePackageDependenciesDesignTime;FindReferenceAssembliesForReferences;_GenerateCompileDependencyCache;_ComputeNonExistentFileProperty;BeforeBuild;BeforeCompile;CoreCompile">
+
+    <Message Importance="High" Text="Analyzing $(MSBuildProjectFile)"/>
+    <Exec
+        ContinueOnError="true"
+        Command="dotnet fsharp-analyzers --project &quot;$(MSBuildProjectFile)&quot; --analyzers-path &quot;$(PkgG-Research_FSharp_Analyzers)\analyzers\dotnet\fs&quot; --exclude-analyzer PartialAppAnalyzer --fail-on-warnings GRA-STRING-001 GRA-STRING-002 GRA-STRING-003 GRA-UNIONCASE-001 --verbose --report &quot;$(MSBuildProjectName)-analysis.sarif&quot;">
```
nojaf, line 40: Explain the `analyzers/dotnet/fs` part. This is a convention.

docs/content/Getting Started.fsx
```diff
@@ -5,217 +5,124 @@ categoryindex: 1
...
 ```
 
-Note that the assembly name needs to contain `Analyzer` in the name in order for it to be picked up.
-
-Add a reference to the analyzers SDK:
-
-```shell
-dotnet add package FSharp.Analyzers.SDK
-```
+Next, add the `PackageReference` pointing to your favorite analyzers to the `.fsproj` file of the project you want to analyzse:
```
nojaf, line 21: `analyzse` typo I think.

docs/content/Getting Started.fsx
```diff
@@ -5,217 +5,124 @@ categoryindex: 1
...
 
-⚠️ Note: To utilize the analyzers in FsAutoComplete (which is subsequently utilized by Ionide), it is essential to ensure that the SDK version matches correctly.
-
-```shell
-paket add FSharp.Analyzers.SDK
+```xml
+<PackageReference Include="G-Research.FSharp.Analyzers" Version="0.1.6">
```
nojaf, line 24: Maybe add a sentence that at the time of writing only the GR analyzers are known to work with the latest cli tool. Link to https://github.com/g-research/fsharp-analyzers and https://www.nuget.org/packages/G-Research.FSharp.Analyzers. This will help if people just want to play with it.

nojaf (changes requested):
A good start but I want to be really thorough on this one.
It is actually the most important documentation page, so we really want to make sure we don't lose the user when they read this.

docs/content/Getting Started Using.fsx
```diff
@@ -0,0 +1,179 @@
...
+
+```xml
+<PackageReference Include="G-Research.FSharp.Analyzers" Version="0.1.6">
+    <PrivateAssets>all</PrivateAssets>
+    <IncludeAssets>build</IncludeAssets>
+</PackageReference>
+```
+
+At the time of writing, the [G-Research analyzers](https://github.com/g-research/fsharp-analyzers) [package](https://www.nuget.org/packages/G-Research.FSharp.Analyzers) contains the only analyzers compatible with the latest CLI tool.  
+Finally, add a custom MSBuild target to the `.fsproj` file for easy invocation of the analyzer:
```
nojaf, line 31: I think this needs a little bit more build-up.
We use the MSBuild target because the `PkgG-Research_FSharp_Analyzers` is convenient.
Maybe we should first give an example of a command line invocation with a hardcoded path.
And thus point out that it is a bit tricky to find that absolute path on each developer's machine.

docs/content/Getting Started Using.fsx
```diff
@@ -0,0 +1,179 @@
...
+
+At last, you can run the analyzer from the project folder:
+
+```shell
+dotnet msbuild /t:AnalyzeProject
+```
+
+## Using analyzers in a solution
+
+First, we need to add the [fsharp-analyzers](https://www.nuget.org/packages/fsharp-analyzers) dotnet tool to the tool-manifest.
```
nojaf, line 59: This needs an introduction paragraph to explain why we are not going to add that previous target to each individual project file. The convenient way to include a target in all `proj` files in a directory is by using `Directory.Build.targets`.

docs/content/Getting Started Using.fsx
```diff
@@ -0,0 +1,179 @@
...
+```shell
+dotnet tool install fsharp-analyzers
```
nojaf, line 61: I'm not sure if I would repeat this section. It feels a bit noisy if you try the single-project approach first.

docs/content/Getting Started Using.fsx
```diff
@@ -0,0 +1,179 @@
...
+```xml
+<ItemGroup>
+    <PackageReference Include="G-Research.FSharp.Analyzers" Version="0.1.6">
+    <PrivateAssets>all</PrivateAssets>
+    <IncludeAssets>build</IncludeAssets>
+    </PackageReference>
+</ItemGroup>
+```
+
+Add the following custom target to the [Directory.Solution.targets](https://learn.microsoft.com/en-us/visualstudio/msbuild/customize-solution-build?view=vs-2022) file to be able to invoke analysis of the whole solution:
```
nojaf, line 75: I would explain again that now that we have added a target for each `proj` file in one place (via the `Directory.Build.targets`), we now want a good way to invoke it for each `proj`.

The alternative is that we run `dotnet msbuild ./src/Project1/Project1.fsproj /t:AnalyzerProject ; dotnet msbuild ./src/Project2/Project2.fsproj /t:AnalyzerProject ...`.
This again doesn't scale and that is why we want to orchestrate things from the solution level.

docs/content/Getting Started Using.fsx
```diff
@@ -0,0 +1,179 @@
...
+```
+
+## Using analyzers in a solution
+
+First, we need to add the [fsharp-analyzers](https://www.nuget.org/packages/fsharp-analyzers) dotnet tool to the tool-manifest.
+```shell
+dotnet tool install fsharp-analyzers
+```
+
+Next, add the `PackageReference` pointing to your favorite analyzers to the [Directory.Build.props](https://learn.microsoft.com/en-us/visualstudio/msbuild/customize-by-directory?view=vs-2022) file:
```
nojaf, line 64: This is again a way to add that `PackageReference` to all `proj` files.
I would really mention this. We will lose users who are unfamiliar with MSBuild I think.

docs/content/Getting Started Using.fsx
```diff
@@ -0,0 +1,179 @@
...
+            <Output TaskParameter="ExitCode" PropertyName="LastExitCode" />
+        </Exec>
+        <Error Condition="'$(LastExitCode)' == '-2'" Text="Problems were found $(MSBuildProjectFile)" />
+
+    </Target>
+
+</Project>
+```
+
+This is effectively the same as adding a target to each `*.proj` file which exist in subfolders.
```
nojaf, line 115: Ok, sorry, ignore my previous comment. I would move this up before showing the code snippet.
You first want to explain the train of thought and then move to the code.

docs/content/Getting Started Using.fsx
```diff
@@ -0,0 +1,179 @@
...
+    <Target Name="AnalyzeSolution">
+        <MSBuild
+                Projects="@(ProjectsToAnalyze)"
+                Targets="AnalyzeProject" />
+    </Target>
+
+</Project>
+```
+
+Finally, add the following custom target to the [Directory.Build.targets](https://learn.microsoft.com/en-us/visualstudio/msbuild/customize-by-directory?view=vs-2022) file.
```
nojaf, line 93: I would move this before the `Directory.Solution.targets` part.

docs/content/Getting Started Using.fsx
```diff
@@ -0,0 +1,179 @@
...
+
+```shell
+dotnet msbuild /t:AnalyzeSolution
+```
+
+## Project Cracking
+
+If all this seems a bit complex to you, let us explain some inner details to give you a better understanding:  
+The way the analyzers work is that we will programmatically type-check a project and process the results with our analyzers. In order to do this programmatically we need to construct the [FSharpProjectOptions](https://fsharp.github.io/fsharp-compiler-docs/reference/fsharp-compiler-codeanalysis-fsharpprojectoptions.html).  
+This is essentially a type that represents all the fsharp compiler arguments. When using --project, we will use [ProjInfo](https://github.com/ionide/proj-info) to invoke a set of MSBuild targets in the project to perform a design-time build.
```
nojaf, line 128: `--project` (add ticks)

nojaf (commented):
I would restructure things a bit.

PR #135, opened 2023-10-30
document what to do in case of multiple target frameworks

dawedawe:
This is very likely not the end of this story but should help people overcome one of the first obstacles the might face.

### DIFF
--- a/docs/content/Getting Started Using.fsx
+++ b/docs/content/Getting Started Using.fsx
@@ -64,6 +64,14 @@ At last, you can run the analyzer from the project folder:
 dotnet msbuild /t:AnalyzeProject
 ```
 
+In case the output tells you there were no analyzers found and you have multiple targets defined in your `TargetFrameworks` property, you might have run into a situation in which there are multiple `$(PkgG-Research_FSharp_Analyzers)` variables and MSBuild doesn't know which one to use.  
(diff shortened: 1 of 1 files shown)


baronfel:
We should make the analysis task multi-targeting aware to help prevent this.

Essentially, if the task is called with TargetFrameworks nonempty, then the task should be called again with each separate TFM and the outputs collected together in some way. I'll send a PR with an example and we can work through the specifics

baronfel:
It would actually be simpler than KeepAChangelog since the package we're talking about here wouldn't have any DLLs - the short version is you would make a package (let's call it FSharp.Analyzers.Build) with the following layout:

```
<root>
- build
- - FSharp.Analyzers.Build.targets
- buildMultitargeting
- - FSharp.Analyzers.Build.targets
```

and in the `build` targets you'd just put the logic for a single-TFM run of the analyzers (since the `build` logic in NuGet packages is guaranteed by the .NET SDK to only run for a single TFM). then, in the `buildMultitargeting` targets you'd have the 'new' target that knows how to handle invoking analysis for each TFM specified (since the `buildMultitargeting` logic in NuGet packages is guaranteed by the .NET SDK to only run in the 'outer', multi-TFM context.

dawedawe:
Not needed anymore because we have https://github.com/ionide/FSharp.Analyzers.SDK/pull/139

PR #145, opened 2023-11-06
Take the TAST (FSharpImplementationFileContents) as parameter in walkTast

dawedawe:


nojaf:
While we are changing the API, would you also mind extending `WalkCall` and `WalkNewRecord`?
Right now, we are only passing the arguments that we were interested in.
I think it would be better to pass all of them. To be a bit more future-proof.
It is easier to ignore some arguments on the call side than be frustrated that they aren't there.

I would also prefer parameter names:
image

src/FSharp.Analyzers.SDK/TASTCollecting.fsi
```diff
@@ -11,12 +11,27 @@ module TASTCollecting =
         new: unit -> TypedTreeCollectorBase
 
         /// Overwriting this member hooks up a custom operation for a call of a member or function.
-        abstract WalkCall: range -> FSharpMemberOrFunctionOrValue -> FSharpExpr list -> unit
-        default WalkCall: range -> FSharpMemberOrFunctionOrValue -> FSharpExpr list -> unit
+        abstract WalkCall:
+            exprRange: range ->
```
nojaf, line 15: I can't help myself, but could we put the range as the last parameter?
That feels more consistent with the untyped syntax tree.
Totally nit, sorry again!

nojaf (approved):

PR #151, opened 2023-11-09
Print error and exit when project cracking failed

### LINKED ISSUE #126: Provide better DX when project cracking failed
**Describe the bug**
Right now https://github.com/ionide/FSharp.Analyzers.SDK/blob/9d2813b171078b15eb0244238120492464e94204/src/FSharp.Analyzers.Cli/Program.fs#L68 will throw hard when the list of cracked projects is empty.

We should print a better message. 
I think for now it is still fine to exit the tool.

dawedawe:
fixes #126


nojaf (approved):
Thanks!

Issue #84, opened 2023-09-10
Add a `-p` flag to allow passing MSBuild properties through to the MSBuild evaluation portion of checking.


baronfel:
**Is your feature request related to a problem? Please describe.**

MSBuild/.NET CLI flags often change behavior - notably flags like `-c`, `-r`, `--arch`, `--os`, and `-p`. At minimum we need to allow for MSBuild properties to be passed in to allow for checking of AST shapes/compiler defines that actually match what the .NET CLI will compile against.

**Describe the solution you'd like**

Minimum:
* -p|--property flag that can be applied multiple times and accepts MSBuild properties. These are key=value pairs.

Bonus - parity with the .NET CLI:
* -r|--runtime flag for passing through the RuntimeIdentifier MSBuild property
* --arch flag for inferring the OS portion of the RID only and explicitly specifying the architecture part
* --os flag for inferring the architecture portion of the RID only and explicitly specifying the OS part
* -c flag for passing through the Configuration MSBuild property

dawedawe:
I think so.
@baronfel are you still missing something?

baronfel:
Nope, we should be all good to go here :)

PR #152, opened 2023-11-09
Support MSBuild properties on the CLI

dawedawe:
@baronfel Is that roughly what you had in mind in #84 ?

With it we can do things like:
```shell
-p Configuration=Release -p DefineConstants="MYDEF" ...
```

### DIFF
--- a/src/FSharp.Analyzers.Cli/Program.fs
+++ b/src/FSharp.Analyzers.Cli/Program.fs
@@ -472,6 +472,10 @@ let main argv =
     let ignoreFiles = ignoreFiles |> List.map Glob
     let properties = getProperties results
 
+    if Option.isSome fscArgs && not properties.IsEmpty then
+        printError "fsc-args can't be combined with MSBuild properties."
+        exit 1
+
     if verbose then
         properties |> List.iter (fun (k, v) -> printInfo $"Property %s{k}=%s{v}")
 
--- a/src/FSharp.Analyzers.Cli/Program.fs
+++ b/src/FSharp.Analyzers.Cli/Program.fs
@@ -15,6 +15,7 @@ type Arguments =
     | Project of string list
     | Analyzers_Path of string list
     | [<EqualsAssignment; AltCommandLine("-p:"); AltCommandLine("-p")>] Property of string * string
+    | [<Unique; AltCommandLine("-c")>] Configuration of string
     | [<Unique>] Treat_As_Info of string list
     | [<Unique>] Treat_As_Hint of string list
     | [<Unique>] Treat_As_Warning of string list
@@ -31,6 +32,7 @@ type Arguments =
             | Project _ -> "Path to your .fsproj file."
             | Analyzers_Path _ -> "Path to a folder where your analyzers are located."
             | Property _ -> "A key=value pair of an MSBuild property."
+            | Configuration _ -> "The configuration to use, e.g. Debug or Release."
             | Treat_As_Info _ ->
                 "List of analyzer codes that should be treated as severity Info by the tool. Regardless of the original severity."
             | Treat_As_Hint _ ->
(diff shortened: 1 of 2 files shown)


baronfel:
Yep! This would do the base-level support - though MSBuild properties can be passed via `-p` in a number of different formats. You can see the tests we have in the SDK for our own parsing of the various formats [here](https://github.com/dotnet/sdk/blob/ee471d60180fbfb8a447ad4ae32a474aefade8ca/src/Tests/dotnet.Tests/ParserTests/MSBuildArgumentCommandLineParserTests.cs#L44-L57). I think simple key=value pairs are the most common, but automated tools especially love to use the more complex formats.

dawedawe:
Thanks for the pointer. These should all work now with `-p` or `-p:` or `--property`.

dawedawe:
If the following mapping to MSBuild properties is correct, I think this PR is ready for review (not sure about OS):
`-r` maps to `RuntimeIdentifier`
`-a` maps to `Platform`
`--os` maps to `OS`

baronfel:
`--os` and `--arch` map to different components of a RuntimeIdentifier. Take `linux-x64` for example - the `linux` portion is the `OS`, and the `x64` portion is the `arch`. This isn't a hard-and-fast rule, however - the Runtime Identifier `linux-musl-x64` has an `os` of `linux-musl` and an `arch` of `x64`.

The dotnet CLI errors if arch/os are provided at the same time as -r, and it 'fills in' any missing pieces of the final RuntimeIdentifier from arch/os with the arch/os of the current SDK. So a user specifying `--arch arm64` on a linux device would end up passing `-p RuntimeIdentifier=linux-arm64` to the build.

dawedawe:
Ah okay sorry, I had some bigger misunderstanding then.
So I need to find a way how to query programatically the correct string values for os and arch then...

baronfel:
The 'correct' way to get the RID for the currently running SDK is [here](https://github.com/dotnet/sdk/blob/dd380277a95e2a2a833bc40a9f0b547ca36c34b2/src/Cli/dotnet/CommonOptions.cs#L230-L250) - but it is very, very gross.

dawedawe:
So `System.Runtime.InteropServices.RuntimeInformation.RuntimeIdentifier` can be wrong in some cases?

baronfel:
oh, great idea - that should be fine for your needs. I think the SDK has some layering problem where it can't necessarily rely on that.

dawedawe:
Ok, thanks a lot for all your help here!

src/FSharp.Analyzers.Cli/Program.fs
```diff
@@ -418,15 +448,7 @@ let main argv =
...
-    let configuration = results.TryGetResult <@ Configuration @>
-
-    let properties =
-        results.GetResults <@ Property @>
-        |> expandMultiProperties
-        |> fun props ->
-            configuration
-            |> Option.map (fun c -> List.append props [ "Configuration", c ])
-            |> Option.defaultValue props
+    let properties = getProperties results
```
nojaf, line 451: Can properties be combined with `--fsc-args`?

dawedawe: I expect they can be combined but I'll test next week.

nojaf: My initial thinking is that they cannot. Properties would influence the project cracking I assume and we skip that entirely when using `--fsc-args`.

dawedawe: As we discussed, while we could shoehorn some parts from the properties into the fsc-args, it would be against the idea of fsc-args, so let's disallow that.

dawedawe:
With the RuntimeIdentifier being constructed with --arch and --os if given, I hope this is ready now.

nojaf (approved):
Looks good, ship it!

PR #159, opened 2023-11-13
use forward slashes in doc examples

dawedawe:
Use forward slashes in doc examples as they work in windows and unix


nojaf:
I think I got most of these in the other PR.
Thanks for pointing it out though!

PR #172, opened 2023-11-24
Add a description to the Cli tool nuget package

Numpsy:
Quite trivial and i'm not sure what the text should be, it's just that whenever I look at listings on nuget.org that don't have a description set, the default 'Package Description' text sticks out to me - e.g.

image


dawedawe (approved):
Thanks!
image

Issue #33, opened 2021-01-11
Please add License to fsharp-analyzers dotnet tool publication


DamianAtWork:
I work in an environment where .NET packages and tools go through an internal package manager and doesn't allow us to use nuget.org directly (not uncommon). Our policies block unlicensed packages and the scanning for licenses is automated. 

Would it be possibly to include the license information with fsharp-analyzers? The currently published packages don't include licensing info.

cmeeren:
The [current NuGet package](https://www.nuget.org/packages/FSharp.Analyzers.SDK/0.11.0) (published in January 2022) contains a standard MIT license, using `PackageLicenseExpression` as per the [official recommendations](https://learn.microsoft.com/en-us/nuget/create-packages/package-authoring-best-practices#licensing). I assume this issue can be closed.

cmeeren:
Nermind, I didn't read carefully enough. It seems this issue pertains to the [fsharp-analyzers](https://www.nuget.org/packages/fsharp-analyzers) package. It does not currently have a license.

Numpsy:
It looks to me like none of the current nuget packages are showing a license on the nuget.org listing?

/wonders if the condition at https://github.com/ionide/FSharp.Analyzers.SDK/blob/ae263a59ef5b453789ba191cac0105ffedfa7b65/Directory.Build.props#L28C3-L28C3 might be getting resolved before 'IsPackable' is set to true in the projects?

baronfel:
Yes, that's exactly correct - D.B.props happens _very_ early in the ordering of building a project:

* D.B.props
* Props from the SDK(s) you've imported (From the Sdk="..." at the top of the project file)
* the project file itself
* Targets from the SDK(s) you've imported
* D.B.targets

baronfel:
Generally if you have any logic (not just static values) I'd suggest that you use a Directory.Build.targets for that instead - it tends to behave most like users expect MSBuild logic to behave.

Numpsy:
Is there a need for ```PackageLicenseExpression``` to be conditioned in any case? (I didn't think it'd do anything for projects that aren't packaged anyway)

Numpsy:
This should be fixed in the 0.22.0 release

dawedawe:
Yep, @Numpsy is correct. It seems to be fixed now.

PR #173, opened 2023-11-24
Remove the condition on setting PackageLicenseExpression/PackageTags in directory.build.props

Numpsy:
refs https://github.com/ionide/FSharp.Analyzers.SDK/issues/33 where the license seems to not be getting applied to the packages due to the condition.

Alternatively the suggestion in https://github.com/ionide/FSharp.Analyzers.SDK/issues/33#issuecomment-1823079531 to move the logix into a directory.build.targets also seems to work, so I can do whichever one is prefered.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -28,7 +28,7 @@
     <PackageReference Include="Ionide.KeepAChangelog.Tasks" PrivateAssets="All" />
   </ItemGroup>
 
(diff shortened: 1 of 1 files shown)


nojaf:
Moving to `directory.build.targets` would be cleanest but it is not the end of the world that the sample projects get two more properties.
I'm going to take this as is.
Thanks @Numpsy

nojaf (approved):

PR #174, opened 2023-12-05
pretty printing with colors and adding the help uri if present, to the console output

smoothdeveloper:


smoothdeveloper:
image

How it shows in the console, before the PR, there is no link and the whole line is of the color of the rule identifier.

PR #175, opened 2023-12-12
Use `Microsoft.Extensions.Logging` instead of `printf` based logging infrastructure

dawedawe:
@baronfel Would be great, if you could maybe also take a look if this makes sense.


src/FSharp.Analyzers.SDK/FSharp.Analyzers.SDK.Client.fs
```diff
@@ -205,7 +200,11 @@ type Client<'TAttribute, 'TContext when 'TAttribute :> AnalyzerAttribute and 'TC
                             let shouldExclude = excludedAnalyzers.Contains(registeredAnalyzer.Name)
 
                             if shouldExclude then
-                                logger.Verbose $"Excluding %s{registeredAnalyzer.Name} from %s{assembly.FullName}"
+                                logger.LogInformation(
+                                    "Excludings{0} from {1}",
```
nojaf, line 204:
```suggestion
                                    "Excluding {0} from {1}",
```

src/FSharp.Analyzers.Cli/Logging.fs
```diff
@@ -0,0 +1,79 @@
...
+        | LogLevel.Debug -> textWriter.Write("debug: ")
+        | LogLevel.Information -> textWriter.Write("info: ")
+        | LogLevel.Warning -> textWriter.Write("warn: ")
+        | LogLevel.Error -> textWriter.Write("error: ")
+        | LogLevel.Critical -> textWriter.Write("critical: ")
+        | _ -> ()
+
+    member private _.SetColor(textWriter: TextWriter, logLevel: LogLevel) =
+        let color =
+            match logLevel with
```
nojaf, line 58: This seems so weird that this is the way you need to do this.
Can you put in any documentation reference?

TheAngryByrd:
I think the SDK/TestSDK needs to rely on the `Microsoft.Extensions.Logging.Abstractions` while the CLI can use the console logger.

nojaf (approved):
This looks good, but I'd like a second opinion.
Also, we can probably release this as the next minor this week.

dawedawe:
> I think the SDK/TestSDK needs to rely on the `Microsoft.Extensions.Logging.Abstractions` while the CLI can use the console logger.

Oh yeah, thanks @TheAngryByrd !

src/FSharp.Analyzers.Cli/Logging.fs
dawedawe: I think the lack of a color setting API in the `TextWriter` makes this the way to go.

nojaf:
FYI @dawedawe we are doing `main` now.

smoothdeveloper:
hey, wondering if going this direction would pave way of doing cool things like https://github.com/dotnet/msbuild/tree/main/src/MSBuild/TerminalLogger achieved?

baronfel (commented):
This looks great! Maybe consider using something like https://github.com/verticalsoftware/vertical-spectreconsolelogger to get intuitive colorized output for free instead of rolling your own.

baronfel (approved):

dawedawe:
Thanks for the review.
I tried [vertical-spectreconsolelogger](https://github.com/verticalsoftware/vertical-spectreconsolelogger) but would prefer to wait for it to mature some more.

PR #176, opened 2023-12-12
Console prettyprinting

smoothdeveloper:
same as #174 but targeting main branch


smoothdeveloper:
but this would conflict with #175

PR #177, opened 2023-12-13
document the FullName footgun we run into

dawedawe:
As we have been burned by this multiple times...


nojaf (approved):
Great addition, thanks.

PR #180, opened 2023-12-16
Use ilogger colorized prettyprint and analyzer uri

smoothdeveloper:
redo #174/#176 on top of @dawedawe #175 (this PR should only show my commit(s) once #175 is merged I think).

I'm not sure how to check the mode that is not using the command line, but I assume; the integration into MSBuild, so please let me know if adjustments are needed there and how I should invoke it in a custom project.

This does two things:
* only colorizes the log level and analyzer identifier with the log level colouring (red: errors, yellow: warnings, etc.) but not the message (it is generally more readable than swath of red, orange, so long the heading has the color
* adds the link to the analyzer URI if it is there (it is always in cyan, but can be a problem depending if console background is light or not, we can consider dark blue in such case)

image

Regarding how I initially did it and now, I see we use "vanilla .NET string formatting" instead of interpolation or fsharp formatting specifyier which enforces there are matching arguments.

I was wondering if this was important to use this less safe way due to relying on the logging infrastructure changes done in #175 or we can reconsider going through F# format specifiers without losing on the logging infrastructure integration?


src/FSharp.Analyzers.Cli/CustomLogging.fs
```diff
@@ -8,20 +8,59 @@ open Microsoft.Extensions.Logging.Console
...
+    let formatMessageAsAnsiColorizedString (color: ConsoleColor) (message: string) =
+        $"{ansiForegroundEscapeCodeOfColorConsole  color}{message}{ansiForegroundEscapeCodeOfColorConsole initialConsoleColor}"
+
 type CustomOptions() =
     inherit ConsoleFormatterOptions()
 
     /// if true: no LogLevel as prefix, colored output according to LogLevel
     /// if false: LogLevel as prefix, no colored output
     member val UseAnalyzersMsgStyle = false with get, set
+    member x.UseColoredOutput = x.UseAnalyzersMsgStyle
```
dawedawe, line 56: I don't think this member is needed. If we want to make the colored output configurable in the future, we can add it then.

src/FSharp.Analyzers.Cli/Program.fs
```diff
@@ -252,23 +253,32 @@ let printMessages (msgs: AnalyzerMessage list) =
 
     let msgLogger = factory.CreateLogger("")
 
-    msgs
-    |> Seq.iter (fun analyzerMessage ->
-        let m = analyzerMessage.Message
+    let useAnsiColorCodes = true
```
dawedawe, line 256: Same as above, just always do the colored output for now.

dawedawe (changes requested):

src/FSharp.Analyzers.Cli/CustomLogging.fs
lcg-gauthier-segay: Makes sense, removed this one.

smoothdeveloper: Makes sense, removed this one.

src/FSharp.Analyzers.Cli/Program.fs
smoothdeveloper: fixed now, thanks for the review!

smoothdeveloper:
after rebasing all seems consistent (including what fantomas formatted, versus comment), let see if this gets green.

smoothdeveloper:
No hurries, no worries :)

PR #187, opened 2023-12-27
Fill in arg of log

Smaug123:
Trivial one.

### DIFF
--- a/src/FSharp.Analyzers.Cli/Program.fs
+++ b/src/FSharp.Analyzers.Cli/Program.fs
@@ -576,7 +576,7 @@ let main argv =
             | projects, None ->
                 for projPath in projects do
                     if not (File.Exists(projPath)) then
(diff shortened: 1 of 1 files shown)


nojaf (approved):
Thanks!

Numpsy:
On a related note, there are Roslyn analysers about that try to detect these sorts of issues in C# code for at least MEL and Serilog, so maybe that's something useful to port? (not that I could say offhand how much effort there is in porting the message template code)

Smaug123:
The converse ("you passed a function as a logging arg") [is one of the G-Research analysers](https://github.com/G-Research/fsharp-analyzers/blob/250108a5f392a04b1c3c064ea294281cae672a88/src/FSharp.Analyzers/LoggingArgFuncNotFullyAppliedAnalyzer.fs#L28); seems sensible to have this one too!

PR #190, opened 2024-01-01
Fix a typo in the docs

Numpsy:
Just something I noticed when reading the docs.

### DIFF
--- a/docs/content/Getting Started Using.md
+++ b/docs/content/Getting Started Using.md
@@ -119,7 +119,7 @@ dotnet msbuild YourSolution.sln /t:AnalyzeFSharpProject
 
 ### Select specific projects
 
-As we don't want to targt all projects of the solution, we create a second custom MSBuild target that calls the project-specific target for all relevant projects.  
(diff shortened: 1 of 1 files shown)


docs/content/Getting Started Using.md
```diff
@@ -119,7 +119,7 @@ dotnet msbuild YourSolution.sln /t:AnalyzeFSharpProject
 
 ### Select specific projects
 
-As we don't want to targt all projects of the solution, we create a second custom MSBuild target that calls the project-specific target for all relevant projects.  
+As we don't want to target all projects of the solution, we create a second custom MSBuild target that calls the project-specific target for all relevant projects.
```
Numpsy, line 122: Whilst in the area - would 'in the solution' sound better than 'of the solution'?

baronfel: Yeah, that reads better to my American English ears.

dawedawe, line 122:
```suggestion
As we don't want to target all projects in the solution, we create a second custom MSBuild target that calls the project-specific target for all relevant projects.  
```

nojaf (approved):
Thanks

PR #194, opened 2024-01-03
Support include analyzers and make life simpler for editor clients

dawedawe:


src/FSharp.Analyzers.SDK/FSharp.Analyzers.SDK.Client.fs
```diff
@@ -127,17 +127,21 @@ module Client =
...
     let registeredAnalyzers =
         ConcurrentDictionary<string, Client.RegisteredAnalyzer<'TContext> list>()
 
-    new() = Client(Abstractions.NullLogger.Instance, Set.empty)
+    new() = Client(Abstractions.NullLogger.Instance)
+
+    member x.LoadAnalyzers(dir: string) : int * int = x.LoadAnalyzers(dir, None)
 
-    member x.LoadAnalyzers(dir: string) : int * int =
+    member x.LoadAnalyzers(dir: string, excludeInclude: ExcludeInclude option) : int * int =
```
dawedawe, line 144: Taking the filter argument here instead of in the constructor makes life easier for editor clients as they can hold onto a simple client instance and change the filter on the fly while reloading analyzers after a config change.

Directory.Packages.props
```diff
@@ -19,9 +19,9 @@
     <PackageVersion Include="Microsoft.Build.Locator" Version="1.4.1" />
     <!-- Need to update Directory.Build.props DotNet.ReproducibleBuilds.Isolated version when updating this-->
     <PackageVersion Include="DotNet.ReproducibleBuilds" Version="1.1.1" PrivateAssets="All" />
-    <PackageVersion Include="Microsoft.Extensions.Logging" Version="8.0.0" />
-    <PackageVersion Include="Microsoft.Extensions.Logging.Abstractions" Version="8.0.0" />
-    <PackageVersion Include="Microsoft.Extensions.Logging.Console" Version="8.0.0" />
+    <PackageVersion Include="Microsoft.Extensions.Logging" Version="6.0.0" />
```
dawedawe, line 22: mea culpa, should have used that version from the start.

src/FSharp.Analyzers.SDK/FSharp.Analyzers.SDK.Client.fs
```diff
@@ -197,16 +201,30 @@ type Client<'TAttribute, 'TContext when 'TAttribute :> AnalyzerAttribute and 'TC
...
-                                )
-
-                            not shouldExclude
+                            match excludeInclude with
+                            | Some(Exclude excluded) ->
+                                let shouldExclude = excluded.Contains(registeredAnalyzer.Name)
+
+                                if shouldExclude then
+                                    logger.LogInformation(
+                                        "Excluding {0} from {1}",
```
Numpsy, line 210: fwiw, I think it's usually recomended to use named place holder values in logging message templates rather than numbers so you get better information when using structured logs (there is a [Roslyn analyzer](https://learn.microsoft.com/en-gb/dotnet/fundamentals/code-analysis/quality-rules/ca2253) that detects that)

dawedawe: Thanks!

src/FSharp.Analyzers.SDK/FSharp.Analyzers.SDK.Client.fs
nojaf: Should this not be `?excludeInclude: ExcludeInclude`? Then it probably doesn't need to be a separate member.

dawedawe: Thanks, that's nicer.

nojaf (approved):
Looks good!

PR #196, opened 2024-01-04
Add a --include-files flag

dawedawe:
- add a --include-files flag
- add globing support for analyzer filters
- rename filter flags for consistency

### DIFF
--- a/src/FSharp.Analyzers.SDK/FSharp.Analyzers.SDK.fsproj
+++ b/src/FSharp.Analyzers.SDK/FSharp.Analyzers.SDK.fsproj
@@ -24,5 +24,6 @@
     <PackageReference Include="FSharp.Compiler.Service" />
     <PackageReference Include="McMaster.NETCore.Plugins" />
     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" />
+    <PackageReference Include="Glob" />
   </ItemGroup>
 </Project>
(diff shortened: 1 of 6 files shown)


src/FSharp.Analyzers.SDK/FSharp.Analyzers.SDK.fsproj
```diff
@@ -24,5 +24,6 @@
     <PackageReference Include="FSharp.Compiler.Service" />
     <PackageReference Include="McMaster.NETCore.Plugins" />
     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" />
+    <PackageReference Include="Glob" />
```
nojaf, line 27: Do we really want this to be inside the SDK?

dawedawe: It's needed for the new globbing support of analyzers include/exclude lists.
Do you see a potential issue with it?

nojaf: I was thinking of having that only in the cli tool.
Avoid as much as dependencies as possible for the analyzer authors.

dawedawe: Okay, I went with a predicate parameter. That way, editor clients can still use globing if they want.

nojaf (approved):
Looks good!

PR #199, opened 2024-01-26
Add RequireQualifiedAccess to Severity type

Smaug123:
We currently export the DU case `Error`, which clashes with FSharp.Core's `Result.Error`. This change means we never conflict with FSharp.Core. The change is source-incompatible, so it will require a release note.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 ## Unreleased
 
 ### Changed
+* [Add `RequireQualifiedAccess` to `Severity` type](https://github.com/ionide/FSharp.Analyzers.SDK/pull/199) (thanks @Smaug123!)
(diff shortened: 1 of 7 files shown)


nojaf:
Could you check https://github.com/ionide/FSharp.Analyzers.SDK/blob/main/docs/content/Getting%20Started%20Writing.fsx ?
It probably needs a change as well.

Smaug123:
Yep, thanks - I can't run this locally easily because I don't have net7 installed, so I'm using GitHub to do it (sorry).

nojaf:
Hmm, fsdocs-tool should be [net6.0](https://github.com/fsprojects/FSharp.Formatting/blob/29f6ed5e9775c571a63433292f43fc9d1729969b/src/fsdocs-tool/fsdocs-tool.fsproj#L6) though. Perhaps you need 7 for other stuff.

Smaug123:
The specific reason is [this line](https://github.com/ionide/FSharp.Analyzers.SDK/blob/1aa73ad9b8204f0954be1734c203abd194f5ad55/samples/OptionAnalyzer.Test/UnitTests.fs#L16). But I'm happy now that this PR is ready.

nojaf (approved):
Thank you!

PR #200, opened 2024-01-26
Fail the tool if any analysis fails

Smaug123:
Motivation is https://github.com/Smaug123/fsharp-prattparser/actions/runs/7619221208/job/20751901249 , which succeeded despite the output being as follows:

```
info: Running in verbose mode
info: Treat as Hints: []
info: Treat as Info: []
info: Treat as Warning: []
info: Treat as Error: [GRA-IMMUTABLECOLLECTIONEQUALITY-001, GRA-INTERPOLATED-001, GRA-JSONOPTS-001, GRA-LOGARGFUNCFULLAPP-001, GRA-STRING-001, GRA-STRING-002, GRA-STRING-003, GRA-TYPE-ANNOTATE-001, GRA-UNIONCASE-001, GRA-VIRTUALCALL-001]
info: Exclude Files: []
info: Include Files: []
info: Loading analyzers from /home/runner/work/fsharp-prattparser/fsharp-prattparser/.analyzerpackages/g-research.fsharp.analyzers/0.7.0/
info: Registered 12 analyzers from 1 dlls
error: Checking of file /home/runner/work/fsharp-prattparser/fsharp-prattparser/PrattParser/obj/Debug/net8.0/.NETCoreApp,Version=v8.0.AssemblyAttributes.fs aborted
error: Checking of file /home/runner/work/fsharp-prattparser/fsharp-prattparser/PrattParser/obj/Debug/net8.0/PrattParser.AssemblyInfo.fs aborted
error: Checking of file /home/runner/work/fsharp-prattparser/fsharp-prattparser/PrattParser/obj/Debug/net8.0/PrattParser.Version.fs aborted
error: Checking of file /home/runner/work/fsharp-prattparser/fsharp-prattparser/PrattParser/Parser.fs aborted
info: No messages found from the analyzer(s)
```

In order not to have a heap of primitive types like `_ list option`, I pulled out the failure case into a single-case DU. That way, the semantics are much clearer if you're only looking at the type signatures.

### DIFF
--- a/src/FSharp.Analyzers.Cli/Program.fs
+++ b/src/FSharp.Analyzers.Cli/Program.fs
@@ -12,6 +12,7 @@ open Microsoft.CodeAnalysis.Sarif
 open Microsoft.CodeAnalysis.Sarif.Writers
 open Microsoft.Extensions.Logging
 open Ionide.ProjInfo
+open FSharp.Analyzers.Cli
 open FSharp.Analyzers.Cli.CustomLogging
 
 type Arguments =
@@ -136,6 +137,7 @@ let runProjectAux
     (fsharpOptions: FSharpProjectOptions)
     (excludeIncludeFiles: Choice<Glob list, Glob list>)
     (mappings: SeverityMappings)
+    : Async<Result<AnalyzerMessage list, AnalysisFailure> list>
     =
     async {
         let! checkProjectResults = fcs.ParseAndCheckProject(fsharpOptions)
@@ -157,25 +159,33 @@ let runProjectAux
                         true
                     | None -> false
             )
-            |> Array.choose (fun fileName ->
+            |> Array.map (fun fileName ->
                 let fileContent = File.ReadAllText fileName
                 let sourceText = SourceText.ofString fileContent
 
                 Utils.typeCheckFile fcs logger fsharpOptions fileName (Utils.SourceOfSource.SourceText sourceText)
-                |> Option.map (Utils.createContext checkProjectResults fileName sourceText)
+                |> Result.map (Utils.createContext checkProjectResults fileName sourceText)
             )
             |> Array.map (fun ctx ->
-                logger.LogInformation("Running analyzers for {0}", ctx.FileName)
-                client.RunAnalyzers ctx
+                match ctx with
+                | Error e -> async.Return(Error e)
+                | Ok ctx ->
+                    async {
+                        logger.LogInformation("Running analyzers for {0}", ctx.FileName)
+                        let! results = client.RunAnalyzers ctx
+                        return Ok results
+                    }
             )
             |> Async.Parallel
 
         return
-            [
-                for messages in messagesPerAnalyzer do
-                    let mappedMessages = messages |> List.map (mapMessageToSeverity mappings)
-                    yield! mappedMessages
-            ]
+            messagesPerAnalyzer
+            |> Seq.map (fun messages ->
+                match messages with
+                | Error e -> Error e
+                | Ok messages -> messages |> List.map (mapMessageToSeverity mappings) |> Ok
+            )
+            |> Seq.toList
     }
 
 let runProject
@@ -266,7 +276,7 @@ let printMessages (msgs: AnalyzerMessage list) =
     let msgLogger = factory.CreateLogger("")
 
     msgs
-    |> Seq.iter (fun analyzerMessage ->
+    |> List.iter (fun analyzerMessage ->
         let m = analyzerMessage.Message
 
         msgLogger.Log(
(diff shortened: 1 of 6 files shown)


src/FSharp.Analyzers.Cli/Program.fs
```diff
@@ -276,15 +285,15 @@ let printMessages (msgs: AnalyzerMessage list) =
...
+            m.Severity.ToString(),
             m.Code,
             m.Message
         )
     )
 
     ()
 
-let writeReport (results: AnalyzerMessage list option) (codeRoot: string option) (report: string) =
+let writeReport (results: AnalyzerMessage list) (codeRoot: string option) (report: string) =
```
Smaug123, line 296: This was never actually optional; we always statically knew that this `option` was `Some`. I tightened up the domain by representing that fact in the types.

src/FSharp.Analyzers.Cli/Program.fs
```diff
@@ -157,26 +159,33 @@ let runProjectAux
...
+                    async {
+                        logger.LogInformation("Running analyzers for {0}", ctx.FileName)
+                        let! results = client.RunAnalyzers ctx
+                        return Ok results
+                    }
             )
             |> Async.Parallel
 
         return
-            Some
```
Smaug123, line 174: There was never any chance that this was `None`, so again I tightened up the types to get rid of the `option`. Instead it's now applied where it's needed (namely in the big `match` in `Program.fs`).

src/FSharp.Analyzers.Cli/Program.fs
```diff
@@ -157,26 +159,33 @@ let runProjectAux
...
                 let sourceText = SourceText.ofString fileContent
 
                 Utils.typeCheckFile fcs logger fsharpOptions fileName (Utils.SourceOfSource.SourceText sourceText)
-                |> Option.map (Utils.createContext checkProjectResults fileName sourceText)
+                |> Result.map (Utils.createContext checkProjectResults fileName sourceText)
             )
             |> Array.map (fun ctx ->
-                logger.LogInformation("Running analyzers for {0}", ctx.FileName)
-                client.RunAnalyzers ctx
+                match ctx with
```
Smaug123, line 170: This could all have been done point-free but I thought it was clearer this way.

src/FSharp.Analyzers.Cli/Program.fs
```diff
@@ -157,26 +159,33 @@ let runProjectAux
...
             |> Async.Parallel
 
         return
-            Some
-                [
-                    for messages in messagesPerAnalyzer do
-                        let mappedMessages = messages |> List.map (mapMessageToSeverity mappings)
-                        yield! mappedMessages
-                ]
+            messagesPerAnalyzer
```
Smaug123, line 182: (Again, this could have been point-free.)

src/FSharp.Analyzers.Cli/Program.fs
```diff
@@ -375,20 +384,6 @@ let writeReport (results: AnalyzerMessage list option) (codeRoot: string option)
         logger.LogError(ex, "Could not write sarif to {report}", report)
         logger.LogInformation("{0}", ex)
 
-let calculateExitCode (msgs: AnalyzerMessage list option) : int =
```
Smaug123, line 378: This is now extracted to the bottom of the file, where the logic is inlined. That's because our exit code now depends on knowledge we only have down there, and I didn't really fancy passing more args into this function.

src/FSharp.Analyzers.Cli/Program.fs
```diff
@@ -623,6 +618,7 @@ let main argv =
             | [], Some fscArgs ->
                 runFscArgs client fscArgs exclInclFiles severityMapping
                 |> Async.RunSynchronously
+                |> Some
```
Smaug123, line 621: (Fallout from the "remove the spurious option type" in FSharp.Analyzers.SDK.fs)

nojaf (commented):
Looks good from afar! I'll let @dawedawe do a proper review as he is more familiar with the latest of the CLI tool (I think).

One thing that would be great if we had some additional help text about the exit codes. On how to interpret them.

Smaug123:
Personally I'm less concerned about documenting exit codes because we already log on all these error cases.

dawedawe:
> Personally I'm less concerned about documenting exit codes because we already log on all these error cases.

Thanks for the work. We can document the exit codes in another PR, then.

dawedawe (approved):

nojaf:
@Smaug123 would you like a new release with all your latest changes?

Smaug123:
That would be great, thanks!

Smaug123:
Thanks - working perfectly (https://github.com/Smaug123/fsharp-prattparser/actions/runs/7708627409/job/21008123060?pr=2 now loudly fails rather than silently failing).

PR #202, opened 2024-01-30
Fail on no analyzers registered

Smaug123:
(I seem to be hitting every possible way to hold this tool wrongly!)

I think it's more useful to fail when there are no analyzers registered, than to return "OK, did nothing". I suspect it's going to be *very* unusual to intentionally register no analyzers.

### DIFF
--- a/src/FSharp.Analyzers.Cli/Program.fs
+++ b/src/FSharp.Analyzers.Cli/Program.fs
@@ -605,7 +605,7 @@ let main argv =
 
     let results =
         if analyzers = 0 then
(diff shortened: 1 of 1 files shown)


nojaf (approved):
Hmm, yes that makes sense.
Thanks for keeping up with us 😇

nojaf:
@dawedawe if you agree, can you approve and merge?

dawedawe (approved):

nojaf:
Sweet 😌

PR #204, opened 2024-01-30
Error when analyzer globs go unmatched

### LINKED ISSUE #203: Have Include/ExcludeFilter warn if an input glob didn't match any analyzers
How would you feel about having IncludeFilter and ExcludeFilter be `Glob list`, instead of `string -> bool`? The FSharp.Analyzers.Cli consumer in this repo does this:

```fsharp
fun (s: string) -> i |> List.map Glob |> List.exists (fun g -> g.IsMatch s)
|> IncludeFilter
```

But what I would *really* like is to be able to report "this particular glob was never matched - are you sure you got it right?". (Doing this would require running every glob against every analyzer, but there's presumably not going to be more than 10k analyzers ever, so that isn't going to be too much of a slowdown.)

* The cleanest way to do this would be to defunctionalise `{Include/Exclude}Filter` to hold only a list of globs. That way, the analyzer SDK could tally how many times it matched each glob against an analyzer, and report the result. Disadvantages: breaking change which possibly removes functionality downstream consumers were depending on; another dependency (to obtain `Glob`)
* An alternative way would be for FSharp.Analyzers.Cli to pass a function which mutates some state, and then check the state at the end. I find that aesthetically displeasing, but it would confine the changes to the CLI rather than requiring a change in the library.

Which of those two would you prefer, if either? (I'm happy to implement either.)

Smaug123:
This is a perf hit, of course, but I claim it's not an important one.

I took the liberty of unindenting the code at the end of the main function (by early-`exit`ing), but this has made the diff pretty nasty, sorry :( 

Fixes #203.

I could understand not wanting this; on balance, though, I think these cases are more likely to be user error than intended.

### DIFF
--- a/src/FSharp.Analyzers.Cli/Program.fs
+++ b/src/FSharp.Analyzers.Cli/Program.fs
@@ -661,6 +661,25 @@ let main argv =
                 |> List.concat
                 |> Some
 
+    let unmatchedAnalyzerPatterns =
+        unmatchedAnalyzerPatterns
+        |> Seq.choose (fun (glob, isUnmatched) -> if isUnmatched then Some glob else None)
+        |> Seq.toList
+
+    // Before we process the results, unconditionally warn if the inclusion or exclusion globs didn't look right.
+    // Then, *after* we've printed the results (which may be useful even if they're partial!), we'll fail if they
+    // didn't look right.
+    let anyUnmatchedAnalyzers =
+        if not (List.isEmpty unmatchedAnalyzerPatterns) then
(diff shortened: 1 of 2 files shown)


nojaf (approved):
This looks good to me. I don't mind the unindent.
I'll let @dawedawe have the final say.

nojaf:
Hi @Smaug123, did you see @dawedawe's last message here?

Smaug123:
I don't really want this feature enough to work on it any more, thanks!

PR #208, opened 2024-02-18
When printing analysis messages to the console, separate the texttual…

Numpsy:
… description from the rest with a ':' instead of '-'

Context: 
I've been trying to get the analyzer results integrated into a CI build at work, which is running in Azure DevOps.

I've got the SARIF results file being consumed by the MS SAST scans tab DevOps extension, and that appears to be working nicely.
However, I though it might be nice if the results could be displayed as regular build warnings as well as in the sarif, so they just display in the main build output instead of you having to look at a different tab to see issues.

I couldn't get the results to show up as warnings, but then noticed that the analysers CLI tool currently outputs the results to the console with the final field separated from the rest with a ```-``` rather than the usual ```:``` and it does seem that changing that causes the output to be recognised as warnings (complete with yellow text output rather than white).

Is this a reasonable change to make?

### DIFF
--- a/src/FSharp.Analyzers.Cli/Program.fs
+++ b/src/FSharp.Analyzers.Cli/Program.fs
@@ -281,7 +281,7 @@ let printMessages (msgs: AnalyzerMessage list) =
 
         msgLogger.Log(
             severityToLogLevel[m.Severity],
-            "{0}({1},{2}): {3} {4} - {5}",
+            "{0}({1},{2}): {3} {4} : {5}",
             m.Range.FileName,
             m.Range.StartLine,
             m.Range.StartColumn,


Numpsy:
Note: I haven't been able to run the unit tests locally due to the issue I mentioned at https://github.com/ionide/FSharp.Analyzers.SDK/pull/182#issuecomment-1951337520

baronfel:
Would you mind adding a screen shot of the before and after that you see in your CI?

Numpsy:
If I run my fun.build script with the current analyzers sdk version, I get this:

image

and with the modification I get this:

image

Also If I enable a binlog in the build, I now get warnings shown in the log:

image

whereas before I only had messages:

image

refs this 'determine what's a warning by pattern matching' behaviour in MSBuild I think? https://learn.microsoft.com/en-us/visualstudio/msbuild/msbuild-diagnostic-format-for-tasks?view=vs-2022

baronfel:
perfect, that's what I expected the cause to be - the SDK is now emitting errors matching the MSBuild Canonical Error Format.

PR #210, opened 2024-05-14
Performance tweaks for larger solutions

TheAngryByrd:
Some performance tweaks for larger solutions when using the CLI tool.

This took my work solution from 15+ minutes to about 2 minutes.

PR #211, opened 2024-05-15
Update MSBuild.StructuredLogger and some MSBuild packages.

nojaf:
Hoping to unblock https://github.com/ionide/FSharp.Analyzers.SDK/pull/210


nojaf:
Ok, I'm not going to claim I fully grasp why but bumping the MSBuild.StructuredLogger now requires the test projects to be `net7.0`. I don't think this is the end of the world.
Would just bump it together with the next FCS release.

PR #213, opened 2024-05-15
update FCS and FSharp.Core for the new release

baronfel:


baronfel:
this needs to wait for proj info before being merged

Issue #215, opened 2024-07-21
Console colour not being reset before fsharp-analyzers cli tool exits?


Numpsy:
When testing #214 I got this in my console (Windows 10 with windows terminal):

![image](https://github.com/user-attachments/assets/0382cba0-4b8d-479e-8dd1-fc42f73b761d)

i.e. it printed the analyzer suggestion in blue, and then left the console colour set to blue such that the next prompt was blue.

I wonder whether https://github.com/ionide/FSharp.Analyzers.SDK/blob/bee7ea7a44b20b09d594dc9e23b407b1f1d086f6/src/FSharp.Analyzers.Cli/CustomLogging.fs#L75 should be writing out the escape codes to reset the colour rather than just setting Console.ForegroundColor ?

Maybe it'd also be fixed by #180 though, I haven't tested it.

dawedawe:
Hey @Numpsy,
thanks for making us aware of this. I'll try to take a look at the end of the week.

PR #216, opened 2024-07-27
Reset the console foreground colour with control codes

Numpsy:
refs #215 

Printing the 0m control code looks like how Fargo resets the colour, and that seems to work for me here on Windows (cmd.exe and windows terminal).
I haven't tested it on any other OS yet, and not sure if there is a need to reset Console.ForegroundColor as well as printing the codes.


dawedawe (approved):
Happens on Linux, too. I wonder if this is a regression from us or from something else as it's pretty obvious :shrug: 
Anyway, thanks a lot for the fix.

dawedawe:
@Numpsy Could you add a changelog entry, please? Then it's good to merge.

Numpsy:
added

dawedawe:
Thanks!

PR #224, opened 2024-11-20
Update Microsoft.Extensions.Logging.Console to 6.0.1

Numpsy:
When I was doing https://github.com/G-Research/fsharp-analyzers/pull/85 I got a security warning from Visual Studio about a transitive dependency

FSharp.Analyzers.SDK.Testing -> Microsoft.Extensions.Logging.Console -> System.Text.Json 6.0.0

Because System.Text.Json 6.0.0 has known CVEs.

So I wondered if it would be useful to update Microsoft.Extensions.Logging.Console to 6.0.1, which has removed the NuGet dependency on System.Text.Json in the .NET 6.0 TFM so it just uses the in-box version.

This shouldn't effect the version of Microsoft.Extensions.Logging.Abstractions used by the SDK lib itself, just the console lib used by the Testing and Cli packages.


nojaf (approved):
Works for me!

PR #225, opened 2025-01-14
Update Getting Started Docs: Calling Analyzers w/ FAKE

1eyewonder:
This PR is a documentation suggestion on calling F# analyzers from FAKE if MSBuild is not the desired workflow a developer wants to take in their repo. Let me know if this is desired or not and if I can make any updates which we think read and/or navigate easier.

An example of this working in the wild can be found here in [FsToolkit.ErrorHandling](https://github.com/demystifyfp/FsToolkit.ErrorHandling/blob/master/build/build.fs#L148-L173)


docs/content/Getting Started Using.md
```diff
@@ -41,6 +41,8 @@ dotnet fsharp-analyzers --project ./YourProject.fsproj --analyzers-path C:\Users
 As you can see, the path to the analyzer DLL files could be tricky to get right across a wide range of setups.  
 Luckily, we can use an MSBuild custom target to take care of the path construction.  
 
+Note: If using MSBuild is not the right solution for you, you can also [call the analyzers from a FAKE build script](#Call-Analyzers-in-Your-FAKE-Build).
```
TheAngryByrd, line 44: I'd just say something simple like:

```suggestion
Note: If you're using Fake you can [call the analyzers from a FAKE build script](#Call-Analyzers-in-Your-FAKE-Build).
```

nojaf:
Feels a little more like a guide on how to call cmd line applications in FAKE.
There is nothing really analyzer specific about any of this.

voronoipotato:
I didn't use this because the msbuild changes seemed confusing and made me afraid that I might make a mistake and mess up the build, @1eyewonder showed me that I could use it with FAKE and I told him he should add something to the documentation here. This approach is a lot less magical to me.

nojaf:
We might consider splitting our current "Getting Started" page into multiple pages. 

First, we could create a dedicated page for all things related to MSBuild. 

Additionally, it would be beneficial to have a page focused on Paket, as it is the most interesting aspect. The specific implementation, whether it's through FAKE or Fun.Build, may not be as important to discuss.

1eyewonder:
So here is what I was thinking for the outline of breaking the docs into multiple pages. Let me know your thoughts

# Getting Started

## Installing Analyzers
Nuget/Paket stuff here

## Configuring Analyzers for Your IDE
I figured it might be more clear if we were to add how to setup up analyzers for VSCode here rather than have users read through an old blog post

## Analyzer Command Line Args
I was thinking maybe adding some docs here about required/optional command line args here if we thought that was useful

## Using MSBuild

## Using FAKE

nojaf:
Assuming each heading is its own page, I'd say this looks reasonable.
Command line args is a tricky one, I would prefer a better `--help` if what we have today is insufficient.

docs/content/getting-started/CLI.md
```diff
@@ -0,0 +1,33 @@
...
+category: getting-started
+categoryindex: 1
+index: 3
+---
+
+# Command Line Arguments
+
+## Example Command
+
+When running the CLI tool from the command line, the bare minimum you need to provide is the path to the project file(s) you want to analyze.
```
nojaf, line 11: This can be a bit misleading. It sounds like something will happen simply by downloading the tool and pointing it to a project, but you actually need to provide an analyzer.

The default argument will only have an effect if you:
- Use Paket in the old mode with that exact default folder name.
- Already have a compatible analyzer.

This scenario seems quite unlikely at this point. It might be better to mention later that if you omit the `--analyzers-path`, it defaults to `packages/analyzers`.

docs/content/getting-started/CLI.md
```diff
@@ -0,0 +1,33 @@
...
+dotnet fsharp-analyzers --project ./YourProject.fsproj
+```
+
+An optional argument you may need to provide is `--analyzers-path`. This is the path to the directory containing the analyzer DLLs. 
+
+```shell
+dotnet fsharp-analyzers --project ./YourProject.fsproj --analyzers-path ./path/to/analyzers/directory
+```
+
+⚠️ If you don't provide this argument, it will default to `packages/analyzers`.
```
nojaf, line 23: This is really a Paket thing and should be mentioned as such. Folks using NuGet might not get how you can ever download the analyzers in that folder.

docs/content/getting-started/Configuring for IDE.md
```diff
@@ -0,0 +1,26 @@
...
+category: getting-started
+categoryindex: 1
+index: 2
+---
+
+# Configuring for the IDE
+
+## Visual Studio Code
+
+In order to configure analyzers for VSCode, you will need to update your project's `.vscode/settings.json` file or your user settings. You should only need the following settings:
```
nojaf, line 11: @TheAngryByrd, where can people find the log if there's an SDK mismatch? Users may use older analyzers, which might not be detected by the tooling. I believe we mention this somewhere. I suggest including an explanation on this page about where this information will be displayed.

docs/content/getting-started/Installing Analyzers.md
```diff
@@ -0,0 +1,50 @@
...
+index: 1
+---
+# Installation
+
+## Installing the Tool
+
+A dotnet CLI tool, called [fsharp-analyzers](https://github.com/ionide/FSharp.Analyzers.SDK/), is used to run analyzers outside the context of an IDE. Add it to your tool-manifest with:
+
+```shell
+dotnet tool install fsharp-analyzers
```
nojaf, line 13:
```suggestion
dotnet tool install fsharp-analyzers --create-manifest-if-needed
```

nojaf (commented):
Great start!

docs/content/getting-started/Configuring for IDE.md
TheAngryByrd: Briefly looking at the code, I'm not sure we actually log this correctly.

Added issue here: https://github.com/ionide/FsAutoComplete/issues/1350

nojaf: Alright, thanks, @1eyewonder maybe we just link to that issue in the docs.

1eyewonder:
Pushed the latest doc changes. Sorry I've been away for a bit. Question regarding the CI build. Is this something that should be breaking if I didn't change anything? I can try to fix this if needed

nojaf (approved):

nojaf:
Thanks for the latest changes, looks good now!
As for the CI, the problem is not related to your changes but could you change https://github.com/ionide/FSharp.Analyzers.SDK/blob/ac7d4751f02026b595dc48de66ebf154b8af8495/samples/OptionAnalyzer.Test/UnitTests.fs#L16 to `net8.0`.
I've seen this in other projects as well.

nojaf:
Thanks, @TheAngryByrd feel free to merge when this is fine for you.

TheAngryByrd (approved):

PR #228, opened 2025-02-26
Prevent File Locking Issue

1eyewonder:
## Problem

In VSCode I currently have the following settings to enable analyzers in the IDE.

```json
{
    "FSharp.enableAnalyzers": true,
    "FSharp.analyzersPath": [
        "packages/Analyzers"
    ],
}
```

When switching between branches of my codebase, if there are different versions of my analyzers, I will get an error during the paket restore step due to the files being used. My current workaround is to manually delete the folder listed above and then restore my solution.

## Proposed Solution

In order to circumvent this, there appears to be a setting in the package we are consuming to allow loading the assemblies into memory. Looking through the code, it does not appear we are using this setting nor is it overriding any hot reload functionality.

[CreateFromAssemblyFile](https://github.com/natemcmaster/DotNetCorePlugins/blob/a6094157bfd1ace9d6b13df108c488526cfe850f/src/Plugins/PluginLoader.cs#L143-L153)

[PluginConfig constructor](https://github.com/natemcmaster/DotNetCorePlugins/blob/a6094157bfd1ace9d6b13df108c488526cfe850f/src/Plugins/PluginConfig.cs#L21-L34)

### DIFF
--- a/src/FSharp.Analyzers.SDK/FSharp.Analyzers.SDK.Client.fs
+++ b/src/FSharp.Analyzers.SDK/FSharp.Analyzers.SDK.Client.fs
@@ -169,6 +169,7 @@ type Client<'TAttribute, 'TContext when 'TAttribute :> AnalyzerAttribute and 'TC
                                 fun config ->
                                     config.DefaultContext <- AssemblyLoadContext.Default
                                     config.PreferSharedTypes <- true
+                                    config.LoadInMemory <- true
                             )
 
                         Some(analyzerDll, analyzerLoader.LoadDefaultAssembly())


1eyewonder:
Here is some additional context https://github.com/natemcmaster/DotNetCorePlugins/issues/87

TheAngryByrd (changes requested):
Looks good but I think we wanna test in FSAC to make sure this works as expected.

1eyewonder:
## Local FSAC

- Pointed towards local FSharp.Analyzers.SDK with the proposed changes
    - image
- Removed references to nuget package
    - image

## Delete Script

The below script is what I use to test if a file is locked by Windows

```fsharp
open System.IO

let path = @"packages/analyzers"
Directory.Delete(path, true)
```

## Opening Using Local Repo as Is

- Analyzers exist upon opening
    - image
- Analyzers functional
    - image
- Run the above script
    - image
  
## Opening Using Local Repo & Local FSAC

- Set local fsac dll path
    - image
- Analyzers exist upon opening
    - image
- Analyzers functional
    - image
- Running the script successfully deletes the packages and does not throw an error
    - image

TheAngryByrd (approved):

nojaf (approved):

PR #238, opened 2025-07-13
Move membersToIgnore/exprTypesToIgnore out of visitDeclaration()

Numpsy:
Just a thought when playing with things - looks like creating these sets inside ```visitDeclaration``` creates a substantial number of lists and sets, and as the contents are constant it doesn't seem like it should need to create them more than once?

Running a memory profile of the SDK cli tool against one of my projects got this before:

image

and this after:

image


TheAngryByrd (approved):

TheAngryByrd:
Thanks! I'll merge it when I'm ready to release.

PR #241, opened 2025-07-27
Utilize Fsc args in msbuild analyze if available

TheAngryByrd:
This kind of revitalizes https://github.com/ionide/FSharp.Analyzers.SDK/pull/164, but this takes a slightly different approach.

- Instead of trying to do a DesignTimeBuild within msbuild, it will try to run after a `CoreCompile`/`AfterBuild` target actually runs.
- This uses the `RunAnalyzer` or `RunAnalyzersDuringBuild` (see [Roslyn variables](https://learn.microsoft.com/en-us/visualstudio/code-quality/disable-code-analysis?view=vs-2022#net-framework-projects)) to determine if it should run or not.
- This will skip the `FsharpAnalyzerAfterBuild` target if `CoreCompile` does not produce FscArgs. However someone can use `FSharpAnalyzers_AlwaysRunAfterBuild` to always run after a build, still using Ionide.ProjInfo as before to do design time builds.

### DIFF
--- a/.config/dotnet-tools.json
+++ b/.config/dotnet-tools.json
@@ -6,7 +6,15 @@
       "version": "20.0.0",
       "commands": [
         "fsdocs"
-      ]
+      ],
+      "rollForward": false
+    },
+    "fsharp-analyzers": {
+      "version": "0.32.0",
+      "commands": [
+        "fsharp-analyzers"
+      ],
+      "rollForward": false
     }
   }
 }
--- a/.config/dotnet-tools.json
+++ b/.config/dotnet-tools.json
@@ -6,15 +6,7 @@
       "version": "20.0.0",
       "commands": [
         "fsdocs"
-      ],
-      "rollForward": false
-    },
-    "fsharp-analyzers": {
-      "version": "0.32.0",
-      "commands": [
-        "fsharp-analyzers"
-      ],
-      "rollForward": false
+      ]
     }
   }
 }
--- a/docs/content/getting-started/MSBuild.md
+++ b/docs/content/getting-started/MSBuild.md
@@ -197,7 +197,7 @@ Run `dotnet msbuild YourProject.fsproj /t:Dump` and verify that `CodeRoot` has a
 
 ## Analyze FSharp Projects After Build
 
-If you'd like the analyzer to be ran after a build, you can set `RunAnalyzersDuringBuild` or `RunAnalyzers` to `true` in your project file:
+If you'd like the analyzer to be run after a `dotnet build`, you can set `RunAnalyzersDuringBuild` or `RunAnalyzers` to `true` in your project file:
 
 ```xml
 <PropertyGroup>
--- a/docs/content/getting-started/MSBuild.md
+++ b/docs/content/getting-started/MSBuild.md
@@ -195,4 +195,30 @@ We often add a dummy target to a project to print out some values:
 
 Run `dotnet msbuild YourProject.fsproj /t:Dump` and verify that `CodeRoot` has a value or not.
 
+## Analyze FSharp Projects After Build
+
+If you'd like the analyzer to be ran after a build, you can set `RunAnalyzersDuringBuild` or `RunAnalyzers` to `true` in your project file:
+
+```xml
+<PropertyGroup>
+    <RunAnalyzersDuringBuild>true</RunAnalyzersDuringBuild>
+    <FSharpAnalyzersOtherFlags>similar to previous section</FSharpAnalyzersOtherFlags>
+</PropertyGroup>
+```
+
+This is reusing the [Roslyn Analyzers variables](https://learn.microsoft.com/en-us/visualstudio/code-quality/disable-code-analysis?view=vs-2022#net-framework-projects-1). For brevity, here are the relevant variables:
+
+- `RunAnalyzersDuringBuild` : Controls whether analyzers run at build time.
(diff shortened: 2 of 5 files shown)


.config/dotnet-tools.json
```diff
@@ -6,7 +6,15 @@
...
         "fsdocs"
-      ]
+      ],
+      "rollForward": false
+    },
+    "fsharp-analyzers": {
+      "version": "0.32.0",
+      "commands": [
+        "fsharp-analyzers"
+      ],
```
TheAngryByrd, line 16: For testing this feature out, I had to add this. It would probably be better to have some properties we can override for local dev, similar to how [Myriad does it](https://github.com/MoiraeSoftware/Myriad/blob/680546237bf5a718eda8593a507d9ee2b7771da4/test/Myriad.IntegrationPluginTests/Myriad.IntegrationPluginTests.fsproj#L67-L68)

src/FSharp.Analyzers.Build/build/FSharp.Analyzers.Build.targets
```diff
@@ -1,11 +1,46 @@
...
+    </Target>
+
+    <Target Name="FsharpAnalyzerAfterBuild" DependsOnTargets="_SetupFSharpAnalyzerProjectOptions" AfterTargets="AfterBuild" Condition="'$(RunAnalyzers)' == 'true'">
+        <Error Condition="$(FSharpAnalyzersOtherFlags) == ''" Text="A property FSharpAnalyzersOtherFlags should exists with all the analyzer cli arguments!" />
+        <!-- 
+            Question: Should we only execute this target if FscCommandLineArgs is not empty? 
+            Argument for Incremental builds may skip CoreCompile if no files changed, so we don't want to run the analyzer in that case.
+            There may be a better way to detect if CoreCompile was skipped but this seems to work.
+            And if someone wants to run the analyzer without FscCommandLineArgs, they can run the AnalyzeFSharpProject target directly.
+        -->
```
TheAngryByrd, line 35: @baronfel

src/FSharp.Analyzers.Build/build/FSharp.Analyzers.Build.targets
```diff
@@ -1,11 +1,46 @@
...
         <Exec
                 ContinueOnError="true"
                 IgnoreExitCode="true"
-                Command="dotnet fsharp-analyzers --project &quot;$(MSBuildProjectFile)&quot; $(FSharpAnalyzersOtherFlags)" />
+                Command="dotnet fsharp-analyzers $(_AnalyzerProjectOptions) $(FSharpAnalyzersOtherFlags)" />
     </Target>
 
     <Target Name="AnalyzeFSharpProject" DependsOnTargets="_AnalyzeFSharpProject" />
+
+    <Target Name="_SetupFSharpAnalyzerProjectOptions" BeforeTargets="CoreCompile" Condition="'$(RunAnalyzers)' == 'true'">
```
TheAngryByrd, line 17: This utilizes the `RunAnalyzers` property that C#/Roslyn analyzers utilize. I think it makes sense to just settle on using it.

TheAngryByrd (commented):
This is definitely a first stab at the problem and we should refine this heavily. I wanted to show off something that works.

nojaf:
Wow, this looks amazing! Will play with this when I can!

baronfel:
> This follows incremental builds so it won't run unless CoreCompile produces FscArgs. 

I'm hearing you say that the output Property/Item isn't being set on purely-incremental builds. I'm not sure that is _expected_ behavior - I'd expect that an incrementally-skipped Target would emit its outputs.

TheAngryByrd:
>  I'd expect that an incrementally-skipped Target would emit its outputs.

Yeah it's not really following the Inputs/Outputs semantics, only running conditionally when `FscCommandLineArgs ` gets set.

samples/MsBuildExample/MsBuildExample.fsproj
```diff
@@ -0,0 +1,27 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <Import Project="../../src/FSharp.Analyzers.Build/buildMultitargeting/FSharp.Analyzers.Build.targets" />
+
+  
+<PropertyGroup>
+    <RunAnalyzersDuringBuild>true</RunAnalyzersDuringBuild>
+    <FSharpAnalyzersOtherFlags>--analyzers-path "../../artifacts/bin/OptionAnalyzer/debug" -v d</FSharpAnalyzersOtherFlags>
+    <FSharpAnalyzers_Exe>../../artifacts/bin/FSharp.Analyzers.Cli/debug/FSharp.Analyzers.Cli.dll</FSharpAnalyzers_Exe>
```
nojaf, line 9: There are some msbuild variables with a `_` in them.
Like mostly PascalCase but then a singular underscore.
What is the rationale for that?

docs/content/getting-started/MSBuild.md
```diff
@@ -195,4 +195,30 @@ We often add a dummy target to a project to print out some values:
 
 Run `dotnet msbuild YourProject.fsproj /t:Dump` and verify that `CodeRoot` has a value or not.
 
+## Analyze FSharp Projects After Build
+
+If you'd like the analyzer to be ran after a build, you can set `RunAnalyzersDuringBuild` or `RunAnalyzers` to `true` in your project file:
```
Numpsy, line 200: "to be **run** after a build" maybe?

PR #244, opened 2025-09-04
Update Ionide.KeepAChangelog.Tasks to 0.3.0

Numpsy:
I'm seeing if doing the update fixes instances of https://github.com/ionide/KeepAChangelog/issues/8 when trying to buld various things in Visual Studio.
This is also requiring tweaks to some of the changelog files, as the new parser seems to be more strict about the expected format of the file


Numpsy:
There is discussion over in https://github.com/fsprojects/fantomas/pull/3182 about why the markdown lists were changed to use ```-``` rather than ```*```.

PR #252, opened 2025-09-21
Fix the path to the OptionsAnalyzer binary in the readme.

Numpsy:
The build is using artifact output, so the binaries get put under ./artifacts rather than ./samples


nojaf:
Thanks!

PR #300, opened 2026-06-04
FSC and FSharp.Core update

Thorium:
Fix the issue exposed by CI after the dependabot update.

### DIFF
--- a/samples/OptionAnalyzer/OptionAnalyzer.fsproj
+++ b/samples/OptionAnalyzer/OptionAnalyzer.fsproj
@@ -13,6 +13,7 @@
     </ItemGroup>
 
     <ItemGroup>
+      <PackageReference Include="FSharp.Compiler.Service" />
         <ProjectReference Include="..\..\src\FSharp.Analyzers.SDK\FSharp.Analyzers.SDK.fsproj" />
     </ItemGroup>
 
--- a/samples/OptionAnalyzer/OptionAnalyzer.fsproj
+++ b/samples/OptionAnalyzer/OptionAnalyzer.fsproj
@@ -13,7 +13,6 @@
     </ItemGroup>
 
     <ItemGroup>
-      <PackageReference Include="FSharp.Compiler.Service" />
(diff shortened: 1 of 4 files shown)


nojaf (approved):
Thanks, looks good.
Would you like a release for this?

Thorium:
Yes, because we need to update the FSharp.Core of the FSAutoComplete, and that repository is pinning this as a dependency.

TheAngryByrd (approved):

nojaf:
Okay, please add a `0.37.0` section for https://github.com/ionide/FSharp.Analyzers.SDK/blob/main/CHANGELOG.md
If that is present, I can make the tag and start the release.

samples/OptionAnalyzer/OptionAnalyzer.fsproj
```diff
@@ -13,6 +13,7 @@
     </ItemGroup>
 
     <ItemGroup>
+      <PackageReference Include="FSharp.Compiler.Service" />
```
nojaf: Why is this necessary?

Thorium: That was automatically added by dependabot commit.

nojaf: What happens if you remove it?

Thorium: Let me try... I would expect nothing, because there is already a reference to that package in Directory.Packages.props that are probably stronger than any single fsproj anyway.

Thorium:
Ok, all good to merge/release.

nojaf:
Sorry for the delay, currently on holiday.
Publish did not come through, see https://github.com/ionide/FSharp.Analyzers.SDK/actions/runs/27695240071/job/81916558098
@TheAngryByrd will update the nuget api key.

baronfel:
We should try to move to NuGet.org trusted publishing like we've done for other repos - I'll send a PR for this in a sec.

nojaf:
@Thorium available in https://www.nuget.org/packages/FSharp.Analyzers.SDK/0.37.2

PR #301, opened 2026-06-17
Add NuGet login step for package publishing

baronfel:
This prevents us from having to rotate NuGet publishing keys, as one is retrieved on-demand.  Docs are at https://learn.microsoft.com/en-us/nuget/nuget-org/trusted-publishing#github-actions-setup.

I've added the NUGET_USER secret that points to the trusted publisher configuration. We do similar in FsAutoComplete and a few other repos already: https://github.com/ionide/FsAutoComplete/blob/main/.github/workflows/release.yml#L54-L61

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -39,6 +39,12 @@ jobs:
           version: ${{ steps.tag_name.outputs.current_version }}
           path: ./CHANGELOG.md
 
+      - name: Get NuGet Publishing Key
+        uses: NuGet/login@v1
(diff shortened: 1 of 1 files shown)


nojaf (approved):
Thanks @baronfel!

nojaf:
@baronfel it did not quite work: https://github.com/ionide/FSharp.Analyzers.SDK/actions/runs/27699932382/job/81933270688

Do lack permission but maybe also some environment thing? What did you configure there?

PR #312, opened 2026-08-27
Update the checkout and setup-dotnet actions

Numpsy:
There are warnings in the build about the current versions using the deprecated Node 20, so lets update them

ref

image


Numpsy:
Should dependabot be doing these?

nojaf:
Yes, dependabot can do these things. We should try and set that up.
Thanks for the PR!

nojaf (approved):

PR #322, opened 2026-09-03
Automatically release new version when detected

nojaf:
Same principle as in https://github.com/ionide/ionide-analyzers:
- You add a new version in the CHANGELOG.md file.
- If that PR gets approved and merge to main, we figure out if we need a new release.
- Do Nuget + GitHub release

### DIFF
--- a/build.fsx
+++ b/build.fsx
@@ -1,14 +1,24 @@
 #!/usr/bin/env -S dotnet fsi --
 
 #r "nuget: Fun.Build, 1.1.18"
+#r "nuget: Ionide.KeepAChangelog, 0.1.8"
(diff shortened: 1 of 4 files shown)


build.fsx
```diff
@@ -1,14 +1,24 @@
 #!/usr/bin/env -S dotnet fsi --
 
 #r "nuget: Fun.Build, 1.1.18"
+#r "nuget: Ionide.KeepAChangelog, 0.1.8"
```
Numpsy: Noddy thing, but is there a reason to not use the latest version of KeepAChangeLog?

nojaf: Nope, pure oversight, let me bump that. Thx!

PR #331, opened 2026-09-11
Minor code cleanup: Hoist the property-splitting Regex out of the map.

Thorium:
Hoist the property-splitting Regex out of the map.

### DIFF
--- a/src/FSharp.Analyzers.Cli/MsBuildProperties.fs
+++ b/src/FSharp.Analyzers.Cli/MsBuildProperties.fs
@@ -6,12 +6,12 @@ open System.Text.RegularExpressions
 open Argu
 open Microsoft.Extensions.Logging
 
-/// If multiple MSBuild properties are given in one -p flag like -p:prop1="val1a;val1b;val1c";prop2="1;2;3";prop3=val3
-/// argu will think it means prop1 has the value: "val1a;val1b;val1c";prop2="1;2;3";prop3=val3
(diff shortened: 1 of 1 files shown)


src/FSharp.Analyzers.Cli/MsBuildProperties.fs
```diff
@@ -9,14 +9,16 @@ open Microsoft.Extensions.Logging
 /// If multiple MSBuild properties are given in one -p flag like -p:prop1="val1a;val1b;val1c";prop2="1;2;3";prop3=val3
 /// argu will think it means prop1 has the value: "val1a;val1b;val1c";prop2="1;2;3";prop3=val3
 /// so this function expands the value into multiple key-value properties
+// built once for every property, not per element
```
nojaf, line 9: This comments is a bit confusing, is it not the other way around?
Also, not sure if I am a fan of the mix of `///` and `//`, does this not mess up the xml doc of expandMultiProperties?

nojaf (changes requested):
Thanks, happy to have this but I have a nitpick on the documentation.

src/FSharp.Analyzers.Cli/MsBuildProperties.fs
Thorium: Oops. Just a sec...

nojaf (approved):

nojaf:
Thx!

